PR #175, opened 2013-08-26
Use EntityManagerInterface to allow simple injection of mocked decorator classes

elmariachi111:
I'm currently writing unit tests for controllers that are protected by an OAuth firewall. I mocked the client's entity manager using the `Doctrine\ORM\Decorator\EntityManagerDecorator` helper class. Unfortunately that class derives from `ObjectManager` in the class hierarchy. Changing the constructor signature to accept an `EntityManagerInterface` that's implemented by each and every Doctrine\ORM\EntityManager is helping out since I can use my EntityManagerDecorators to mock and test entity serialization. The reason why this affects the OAuthBundle is that the firewall depends on a valid instance of an EntityManager. I can't mock a complete EntityManager (returning user objects and tokens) so easily so I'm using a best-of-both-worlds approach: decorate the Entitymanager and mock its `persist` method -> OAuth classes work as expected (when no data is written), my Entities are persisted through overridden methods.

### DIFF
--- a/Entity/AuthCodeManager.php
+++ b/Entity/AuthCodeManager.php
@@ -12,6 +12,7 @@
 namespace FOS\OAuthServerBundle\Entity;
 
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use FOS\OAuthServerBundle\Model\AuthCodeInterface;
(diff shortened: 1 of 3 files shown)


gnat42:
Wow this is an old pull request. I just basically created one (#333) since I didn't see it in

Ener-Getick:
Closed in favor of #333.

PR #198, opened 2013-12-02
added login event to firewall listener (like security.interactive_login)

lukey78:
I'm using the OAuthServerBundle with the Password grant to directly login users from a frontend into a backend accessible through an API. 
In the backend I normally set a lastLoginDate through the security.interactive_login event. This event is not thrown when logging in through the OAuth firewall.

I implemented a simple OAuthLogin event that is thrown when a user logs in through the OAuth firewall. The event just contains the token with the authenticated user.

Please comment if there's a better or more general way to do it. If the implementation is right, I'll also submit an appropriate test. Thanks.

### DIFF
--- a/Event/OAuthLoginEvent.php
+++ b/Event/OAuthLoginEvent.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace FOS\OAuthServerBundle\Event;
+
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+
+/**
+ * Class OAuthLoginEvent
+ * @author Jens Hassler <j.hassler@iwf.ch>
+ */
+class OAuthLoginEvent extends Event
+{
+    const LOGIN =  'fos_oauth_server.login';
+
+    private $authenticationToken;
+
+    /**
+     * Constructor.
+     *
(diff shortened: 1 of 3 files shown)


cfuerst:
i would also be interested in this, since i use resource owner password credentials grant for my own applications i have no way on hooking into the login success flow for things like writing last_login etc.

any thoughts on merging this? is anything missing?

Cheers

Spomky:
I agree with the fact that this bundle does not have any event (e.g. creation/deletion/revocation of tokens).
But I disagree with this PR because the OAuth2 framework protocol is an authorization protocol, not an authentication one.
When a client uses ROPC grant type to get an access token, this bundle can not affirm the RO is logged in.
Furthermore to this acknowledgement, even the OpenID Connect protocol, which is an authorization protocole on top of OAuth2, does not use the ROPC grant type to perform authentication.

cfuerst:
So the better approach would be that for the ROPC FOS\OAuthServerBundle\Storage\OAuthStorage::checkUserCredentials() makes more use of the Symfony AuthenticationProviderManager::authenticate() which would also have the advantage of OAuthServerBundle easily could have multiple user providers?

lukey78:
Closing this because too much time passed without any activity or decision. I'm not up2date what happened in the last 2 years to this bundle, therefore I'm not sure what's currently the best way to handle login events with this bundle.

PR #228, opened 2014-04-08
Update documentation -> user provider settings

rcousens:
Current versions of FOSUserBundle warn that using UserManager as a UserProvider is deprecated and wrong!


Ener-Getick:
Thank you @rcousens

PR #288, opened 2014-11-14
Extract overridable method to handle account status exception.

### LINKED ISSUE #287: Disabled users should be able to perform some actions.
I know this is counter-intuitive. But in my case, a disabled user should perform some actions (fill missing profile properties) by API. These should be logged requests as I want to know who performs these profile changes.

I agree that LOCKED users should NOT be granted access, but that is not the case for disabled users.

see answer of @stof [here](https://groups.google.com/forum/#!topic/symfony2/aG79IY7Kmc8)

So merge #277 broke my API.

I will extend the OAuthProvider to be able to take right decision on pre/post `AccountStatusException`.

bburnichon:
Closes #287

Users of the library can choose to allow access based on their
requirements. For example: disabled users can access the API to complete
their profile.

I currently handles this case by changing `fos_oauth_server.security.authentication.provider.class` to my own implementation which replicate the `authenticate` method.


bburnichon:
Sorry @alanbem, I missed your comment. I squashed my commits.

PR #297, opened 2015-02-13
Pass the requested scope with the OAuthEvent on authorization

### LINKED ISSUE #182: Lack of requested scope on pre and post authorization event.
I have created auto-authorization listener following documentation at https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Resources/doc/the_oauth_event_class.md . After that I extended this mechanism with scope validation/check as I didn't want to bypass authorization - of already authorized client - if client requested different scope than before (user should be aware of additional scopes). 

``` php
<?php

// [...]

class OAuthEventListener
{
    // [...]

    public function onPreAuthorizationProcess(OAuthEvent $event)
    {
        if ($user = $this->getUser($event)) {
            if (true === $user->isAuthorizedClient($event->getClient())) {

                $authorizedClient = $user->getAuthorizedClient($event->getClient()); // previously authorized app

                $form = $this->container->get('fos_oauth_server.authorize.form')->getName();
                $scope = $this->container->get('request')->query->get('scope', $this->container->get('request')->request->get($form . '[scope]', '', true));

                /* @var $authorizedClient AuthorizedClient */
                if (true === $authorizedClient->hasIdenticalScope($scope)) {
                    $event->setAuthorizedClient(true);
                }
            }
        }
    }

    public function onPostAuthorizationProcess(OAuthEvent $event)
    {
        if ($event->isAuthorizedClient()) {
            if (null !== $client = $event->getClient()) {

                $form = $this->container->get('fos_oauth_server.authorize.form')->getName();
                $scope = $this->container->get('request')->query->get('scope', $this->container->get('request')->request->get($form . '[scope]', '', true));

                $this->getUser($event)->authorizeClient($client, $scope);
            }
        }
    }

// [...]
}
```

The problem is that I have to pick out 'scope' directly from the request which feels awkward and looks ugly:

``` php
$form = $this->container->get('fos_oauth_server.authorize.form')->getName();
$scope = $this->container->get('request')->query->get('scope', $this->container->get('request')->request->get($form . '[scope]', '', true));
```

IMO scope parameter would be a nice addition to *authorization event.
Also documentation should mention that if currently requested scope is not validated againts previously authorized scope - in the auto-authorization mechanism - it opens the way for possible attack vector (lets call it "scope extortion").

rjelierse:
Pass along the scope that is requested in the authorization process, so listeners to the pre/post authorization events can use it.

Fixes #182

### DIFF
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -52,14 +52,14 @@ class AuthorizeController extends ContainerAware
         $form = $this->container->get('fos_oauth_server.authorize.form');
         $formHandler = $this->container->get('fos_oauth_server.authorize.form.handler');
 
+        $scope = $this->container->get('request')->get('scope', null);
+
         $event = $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::PRE_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient())
+            new OAuthEvent($user, $this->getClient(), false, $scope)
         );
 
         if ($event->isAuthorizedClient()) {
-            $scope = $this->container->get('request')->get('scope', null);
-
             return $this->container
                 ->get('fos_oauth_server.server')
                 ->finishClientAuthorization(true, $user, $request, $scope);
@@ -93,7 +93,7 @@ class AuthorizeController extends ContainerAware
 
         $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::POST_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient(), $formHandler->isAccepted())
+            new OAuthEvent($user, $this->getClient(), $formHandler->isAccepted(), $formHandler->getScope())
         );
 
         $formName = $this->container->get('fos_oauth_server.authorize.form')->getName();
--- a/Event/OAuthEvent.php
+++ b/Event/OAuthEvent.php
@@ -37,13 +37,19 @@ class OAuthEvent extends Event
     private $isAuthorizedClient;
 
     /**
+     * @var string|null
+     */
+    private $scope;
+
+    /**
      * @param UserInterface $user
      */
-    public function __construct(UserInterface $user, ClientInterface $client, $isAuthorizedClient = false)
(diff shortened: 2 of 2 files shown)


Event/OAuthEvent.php
```diff
@@ -37,13 +37,19 @@ class OAuthEvent extends Event
     private $isAuthorizedClient;
 
     /**
+     * @var string
```
Spomky: * @var null|string

Event/OAuthEvent.php
```diff
@@ -37,13 +37,19 @@ class OAuthEvent extends Event
...
+     */
+    private $scope;
+
+    /**
      * @param UserInterface $user
      */
```
Spomky: * @param boolean   $isAuthorizedClient
* @param null|string $scope

Event/OAuthEvent.php
```diff
@@ -77,4 +83,12 @@ public function getClient()
     {
         return $this->client;
     }
+
+    /**
+     * @return string
```
Spomky: * @return null|string

Spomky:
Minor phpdoc to be modified.
@alanbem good for me.

rjelierse:
Would you like me to fix, or will it be taken care of with merge?

allan-simon:
what is the status of this PR ?

rjelierse:
Just updated the PHPDoc for the parts I changed to include possibility of `null` and rebased on top of current master

Controller/AuthorizeController.php
```diff
@@ -52,14 +52,14 @@ public function authorizeAction(Request $request)
         $form = $this->container->get('fos_oauth_server.authorize.form');
         $formHandler = $this->container->get('fos_oauth_server.authorize.form.handler');
 
+        $scope = $this->container->get('request')->get('scope', null);
```
vierbergenlars: This is incomplete. When the form is submitted, `$scope` will be null instead of the scope submitted by the form.
This may pose problems to consumers of `OAuthEvent`, because they suddenly receive an empty scope when the form is submitted. If this influences the resut of `$event->isAuthorizedClient()` so that it becomes `true` while it should have returned `false` for these scopes, this is a security issue. (I tried to fix this #332, where I reimplemented this PR because i did not look down the PR list far enough)

PR #304, opened 2015-04-07
Add unique roles to authentication

nettob:
Add unique roles to token so the roles keep clean!

### DIFF
--- a/Security/Authentication/Provider/OAuthProvider.php
+++ b/Security/Authentication/Provider/OAuthProvider.php
@@ -95,6 +95,8 @@ class OAuthProvider implements AuthenticationProviderInterface
                     }
                 }
 
+                $roles = array_unique($roles);
+
                 $token = new OAuthToken($roles);
                 $token->setAuthenticated(true);
                 $token->setToken($tokenString);


Ener-Getick:
What's your use case?

nettob:
None

Ener-Getick:
I see this is more for performance. Why not :-)

Thank you @nettob

Davidmattei:
Having a problem with this PR, my roles can also be Objects.
So I'm getting this error : **Error: Object of class X could not be converted to string**

Can we fix this by doing : $roles = array_unique($roles, **SORT_REGULAR**);

Greetings

Ener-Getick:
Thank you @Davidmattei I fixed it in https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/commit/0b25cdaae8983c630bb62d14b6993219b1dadb8d

Davidmattei:
Thanks !

nettob:
:+1:

ruscon:
http://php.net/manual/en/function.array-unique.php

```
5.2.9	Added the optional sort_flags defaulting to SORT_REGULAR. Prior to 5.2.9, this function used to sort the array with SORT_STRING internally.
```

Davidmattei:
@ruscon 

`5.2.10	Changed the default value of sort_flags back to SORT_STRING.`

PR #306, opened 2015-04-15
Fix deprecated factory definition

Soullivaneuh:
This PR replace `factory-service` and `factory-method` attributes by `factory` section, according to Symfony deprecate rules.

Had to do this on the extension class instead of xml config to keep Symfony 2.3+ BC.

Regards.


stof:
thanks

PR #308, opened 2015-04-19
Refactor of Random util class with addition of tests

rheh:
Attempted to make the Random class testable within PHPUnit


stof:
Given that I just merged a class bumping the Symfony requirement to 2.2+, we should just use the Symfony implementation

PR #315, opened 2015-04-27
Fix deprecated SecurityContextInterface usage

Soullivaneuh:
This PR fix deprecated SecurityContextInterface usage, introduce since Symfony 2.6.

Symfony 2.3+ BC kept.

@willdurand @stof Have you any suggestion of how handle it easily? Thanks.

### DIFF
--- a/DependencyInjection/Security/Factory/OAuthFactory.php
+++ b/DependencyInjection/Security/Factory/OAuthFactory.php
@@ -36,7 +36,13 @@ class OAuthFactory implements SecurityFactoryInterface
             ;
 
         $listenerId = 'security.authentication.listener.fos_oauth_server.'.$id;
-        $container->setDefinition($listenerId, new DefinitionDecorator('fos_oauth_server.security.authentication.listener'));
+        // Set the SecurityContext for Symfony <2.6
+        $tokenStorageReference = interface_exists('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')
+            ? new Reference('security.token_storage') : new Reference('security.context');
+        $container
+            ->setDefinition($listenerId, new DefinitionDecorator('fos_oauth_server.security.authentication.listener'))
+            ->replaceArgument(0, $tokenStorageReference)
+        ;
 
         return array($providerId, $listenerId, 'fos_oauth_server.security.entry_point');
     }
--- a/Resources/config/security.xml
+++ b/Resources/config/security.xml
@@ -18,7 +18,7 @@
         </service>
 
         <service id="fos_oauth_server.security.authentication.listener" class="%fos_oauth_server.security.authentication.listener.class%" public="false">
-            <argument type="service" id="security.context"/>
+            <argument /> <!-- security.token_storage or security.context for Symfony <2.6 -->
             <argument type="service" id="security.authentication.manager" />
             <argument type="service" id="fos_oauth_server.server" />
         </service>
--- a/Security/Firewall/OAuthListener.php
+++ b/Security/Firewall/OAuthListener.php
@@ -16,6 +16,7 @@ use OAuth2\OAuth2;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
 use Symfony\Component\Security\Core\SecurityContextInterface;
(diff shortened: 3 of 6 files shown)


Soullivaneuh:
Tests added. Ready for review.

Resources/config/security.xml
```diff
@@ -6,6 +6,7 @@
 
     <parameters>
         <parameter key="fos_oauth_server.security.authentication.provider.class">FOS\OAuthServerBundle\Security\Authentication\Provider\OAuthProvider</parameter>
+        <parameter key="fos_oauth_server.security.authentication.listener.legacy.class">FOS\OAuthServerBundle\Security\Firewall\LegacyOAuthListener</parameter>
```
stof: please don't introduce any new class parameters, given that Symfony discourages this usage

Security/Firewall/AbstractOAuthListener.php
```diff
@@ -0,0 +1,86 @@
...
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Http\Firewall\ListenerInterface;
+
+/**
+ * AbstractOAuthListener class.
+ *
+ * @author Arnaud Le Blanc <arnaud.lb@gmail.com>
+ */
+abstract class AbstractOAuthListener implements ListenerInterface
```
stof: please don't use 2 classes. Instead, make the OAuthListener class accept both SecurityContextInterface and TokenStorageInterface as first argument (it is easy given that the way to set the token is the same in both cases anyway)

DependencyInjection/Security/Factory/OAuthFactory.php
```diff
@@ -36,7 +37,11 @@ public function create(ContainerBuilder $container, $id, $config, $userProvider,
             ;
 
         $listenerId = 'security.authentication.listener.fos_oauth_server.'.$id;
-        $container->setDefinition($listenerId, new DefinitionDecorator('fos_oauth_server.security.authentication.listener'));
+        if (version_compare(Kernel::VERSION, '2.6', '>=')) {
```
stof: please use feature detection rather than version detection. When using standalone components, the HttpKernel and the DependencyInjection components might be at different versions

Resources/config/security.xml
Soullivaneuh: Ok I'll put it directly on Symfony definition. I don't know that is discouraged also for external bundles.

Soullivaneuh:
@stof I following your recommendation and take profit to remove all parameters classes on `security.xml` configuration file.

Thanks for your advice, this PR is now ready for review and merge.

chriwup:
@Soullivaneuh did you find out why the tests fail?

Soullivaneuh:
Tests should be fixed. :+1:

Soullivaneuh:
Don't know for the lowest deps failling test, can't reproduce it on my local.

Any idea?

Soullivaneuh:
@stof Could you please restart it?

Soullivaneuh:
Seems to be an issue with php 5.3. Still investigating.

Soullivaneuh:
Can't understand what happen. I got the same error on my local with PHP 5.3 but the ContainerBuilder class **exists**...

sarcher:
I thought perhaps this error was due to the odd namespace (from a different bundle) on the new test. Tried to update this PR with it but looks like I can't do that. Maybe @Soullivaneuh could do that and we'll see if it passes? Would be nice to get this merged in.

Soullivaneuh:
Note: Test @sarcher commit: https://github.com/sarcher/FOSOAuthServerBundle/commit/89d96efaefb82984fdb7182033a75cf02c68d0c4

DependencyInjection/Security/Factory/OAuthFactory.php
```diff
@@ -16,6 +16,7 @@
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\DefinitionDecorator;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\Kernel;
```
stof: this use statement looks unused

Security/Firewall/OAuthListener.php
```diff
@@ -29,34 +30,38 @@
 class OAuthListener implements ListenerInterface
 {
     /**
-     * @var \Symfony\Component\Security\Core\SecurityContextInterface
+     * @var TokenStorageInterface|SecurityContextInterface
      */
-    protected $securityContext;
+    protected $tokenStorage;
```
stof: renaming a protected property is a BC break

Soullivaneuh: I restore property with a deprecated tag. Tell me if it looks good to you.

Soullivaneuh:
@stof Don't know why, but tests fail on php 5.3 becaue of symfony/dependency-injection 2.0.

I added a conflict to bump it to 2.1.

Security/Firewall/OAuthListener.php
```diff
@@ -29,34 +30,47 @@
...
+     *
+     * @deprecated since 1.4.3, to be removed in 2.0
      */
     protected $securityContext;
 
     /**
-     * @var \Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface
+     * @var TokenStorageInterface|SecurityContextInterface
+     */
+    protected $tokenStorage;
```
stof: please make the new property private. Using private visibility makes maintenance much easier, as we don't need to care about BC on them

composer.json
```diff
@@ -29,6 +29,9 @@
         "doctrine/doctrine-bundle": "~1.0",
         "doctrine/orm": ">=2.2,<2.5-dev"
     },
+    "conflict": {
+        "symfony/dependency-injection": "<2.1"
```
stof: I would rather add a dependency on it as we need it anyway

Soullivaneuh: I added it on conflict because it's not a direct dependency but a dependency of framework-bundle.

Will update it.

Security/Firewall/OAuthListener.php
Soullivaneuh: Good idea, will work on it.

BTW, if we don't want override, class should be final later. ;-)

Soullivaneuh:
@stof review fixed.

Soullivaneuh:
I'm sorry but why closing this @Ener-Getick?

Ener-Getick:
I closed it in favor of #374 which fixes all compatibility issues with symfony 3.0 ;-)

Soullivaneuh:
OK great. :+1:

PR #316, opened 2015-04-27
Fix deprecated routing configuration

Soullivaneuh:
This PR fix deprecated routing config throwing related errors:

> The "pattern" is deprecated since version 2.2 and will be removed in 3.0. Use the "path" option in the route definition instead.

> The "_method" requirement is deprecated since version 2.2 and will be removed in 3.0. Use the setMethods() method instead or the "methods" option in the route definition.

We have to bump `symfony/framework-bundle` to `~2.2` version. Not a big deal because 2.1 is not maintained anymore.

The added tests reveal the errors by loading routing configuration files.


stof:
thanks

garak:
Any change to see this in a stable tag?

PR #331, opened 2015-06-25
Do not depend on deprecated 'request' service, removed in symfony 3.0

vierbergenlars:
The 'request' service is deprecated in symfony 2.4, and removed in symfony 3.0.
The alternative is to let the request be injected in the controller, and pass it as a parameter to the methods that require a request.
Another possibility is injecting the 'request_stack' service, but this would not be compatible with symfony < 2.4


Ener-Getick:
Closed in favor of #374

PR #332, opened 2015-06-25
Pass OAuth scopes to the OAuthEvent

vierbergenlars:
An eventlistener on `pre/post_authorization_process` may be interested in the scopes the OAuth client requested. (For example to show the authorization page again when the scopes the app requests have changed since the last authorization)

This PR adds one extra optional parameter to `OAuthEvent`, and thus is fully backwards compatible with previous versions.

### DIFF
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -52,14 +52,14 @@ class AuthorizeController extends ContainerAware
         $form = $this->container->get('fos_oauth_server.authorize.form');
         $formHandler = $this->container->get('fos_oauth_server.authorize.form.handler');
 
+        $scope = $request->query->get('scope', $request->request->get($form->getName().'[scope]', null, true));
+
         $event = $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::PRE_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient())
+            new OAuthEvent($user, $this->getClient(), false, $scope)
         );
 
         if ($event->isAuthorizedClient()) {
-            $scope = $this->container->get('request')->get('scope', null);
-
             return $this->container
                 ->get('fos_oauth_server.server')
                 ->finishClientAuthorization(true, $user, $request, $scope);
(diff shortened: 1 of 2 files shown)


Controller/AuthorizeController.php
```diff
@@ -52,14 +52,14 @@ public function authorizeAction(Request $request)
         $form = $this->container->get('fos_oauth_server.authorize.form');
         $formHandler = $this->container->get('fos_oauth_server.authorize.form.handler');
 
+        $scope = $request->query->get('scope', $request->request->get($form->getName().'[scope]', null, true));
```
Spomky: The way you get scope from the form should not be hardcoded: `$form->getName().'[scope]'`

Event/OAuthEvent.php
```diff
@@ -37,13 +37,19 @@ class OAuthEvent extends Event
...
     /**
+     * @var string
+     */
+    private $scopes;
+
+    /**
      * @param UserInterface $user
      */
-    public function __construct(UserInterface $user, ClientInterface $client, $isAuthorizedClient = false)
+    public function __construct(UserInterface $user, ClientInterface $client, $isAuthorizedClient = false, $scopes = '')
```
Spomky: With the call at line L96, scopes could be null or a string.
So `$scopes = ''` should be `$scopes = null` also. Or you can change the default value at line L96.

Event/OAuthEvent.php
```diff
@@ -77,4 +83,12 @@ public function getClient()
     {
         return $this->client;
     }
+
+    /**
+     * @return string
```
Spomky: * @return string|null

Controller/AuthorizeController.php
vierbergenlars: The way [`AuthorizeController::processSuccess()`](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Controller/AuthorizeController.php#L100-L102) works is by putting all fields userneath the forms namespace in `$request->query`. The OAuth library used requires `scope` to have that parameter name, so making it dynamic seems illogical.

However, I could use `$form->get('scope')->getData()` instead of reading from the request directly. Is that more in line with your expectations? The field name actually is still hardcoded that way.

PR #333, opened 2015-06-30
Use the EntityManager interface ObjectManager

gnat42:
Its possible to decorate the EntityManager which will result in all sorts of type errors. Using the ObjectManager interface allows for decorated EntityManagers


gnat42:
Seems the build failures are unrelated to the change and completely related to some composer/networking issue.

Ener-Getick:
Thank you @gnat42

PR #336, opened 2015-07-14
Defer getRepository

garak:
Method ``getRepository()`` in the constructor is called every time the container is built, even if the service is not used.

### DIFF
--- a/Entity/AuthCodeManager.php
+++ b/Entity/AuthCodeManager.php
@@ -23,11 +23,6 @@ class AuthCodeManager extends BaseAuthCodeManager
     protected $em;
 
     /**
-     * @var \Doctrine\ORM\EntityRepository
-     */
-    protected $repository;
-
-    /**
      * @var string
      */
     protected $class;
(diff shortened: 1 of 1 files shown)


Spomky:
Hi,

You are right this part could be optimized, but with your solution the method `getRepository()` could be called many times.
Hereafter another solution:
```php
/**
 * @var \Doctrine\ORM\EntityRepository|null
 */
protected $repository = null;

/**
 * @return \Doctrine\ORM\EntityRepository
 */
public function getRepository()
{
    if(null === $this->repository) {
        $this->repository = $this->em->getRepository($this->class);
    }
    return $this->repository;
}

public function findAuthCodeBy(array $criteria)
{
    return $this->getRepository()->findOneBy($criteria);
}
```

garak:
As you can see in Doctrine EntityManager, ``createRepository()`` calls a factory. So, there's no need to store it somewhere, you can just call it any time and get the same object.

Spomky:
OK it makes sense.

garak:
Let me link [this article](https://tideways.io/profiler/blog/5-ways-to-optimize-symfony-baseline-performance) of @beberlei in which avoiding Doctrine repositories in constructor is **first** point.

beberlei:
@garak The points are not ordered by severity though :) This change makes sense though

garak:
I know it's not really nice bumping up PRs, anyway this one is very simple and can provide a performance gain. Can you consider to merge it? Thanks.

Ener-Getick:
Thank you @garak

PR #339, opened 2015-08-07
Check if user is blocked

GendelfLugansk:
If account is blocked (http://symfony.com/doc/current/cookbook/security/entity_provider.html#security-advanced-user-interface), user should not be authorized


rtuin:
Nice proposal. What should be done to get this released?

Storage/OAuthStorage.php
```diff
@@ -163,6 +168,24 @@ public function checkUserCredentials(IOAuth2Client $client, $username, $password
...
+            }
+
+            if (!$user->isCredentialsNonExpired()) {
+                throw new CredentialsExpiredException('Credentials is expired');
+            }
+
+            if (!$user->isEnabled()) {
+                throw new DisabledException('Account is disabled');
+            }
+        }
```
rtuin: I think it's best to do these checks after the username/password has been checked. This makes it harder for clients to fish for existing usernames.

PR #342, opened 2015-08-11
Use parameters to define class for service

borNfreee:
It will allow to override class that should be used for this service and will allow to extend functionality without creating CompilerPass.

### DIFF
--- a/Resources/config/oauth.xml
+++ b/Resources/config/oauth.xml
@@ -6,6 +6,7 @@
 
     <parameters>
         <parameter key="fos_oauth_server.server.class">OAuth2\OAuth2</parameter>
+        <parameter key="fos_oauth_server.controller.token.class">FOS\OAuthServerBundle\Controller\TokenController1</parameter>
     </parameters>
 
     <services>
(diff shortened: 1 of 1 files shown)


mvrhov:
Well the symfony doesn't recommend this anymore. I know it's easier to do with just parameter replacement.

stof:
-1 for this

borNfreee:
> symfony doesn't recommend this anymore

@mvrhov where I can read about this and what symfony recommends now, I didn't find after some googling?
@stof if you will have a chance, could you please explain why -1 or point me to where to read why this is a bad to use parameters for classes?

It is interesting because there is already one paramater for other class in this file.

borNfreee:
Found: http://symfony.com/doc/current/best_practices/business-logic.html
>Don't define parameters for the classes of your services.

> This practice was wrongly adopted from third-party bundles. When Symfony introduced its service container, some developers used this technique to easily allow overriding services. However, overriding a service by just changing its class name is a very rare use case because, frequently, the new service has different constructor arguments.

But still don't understand why *this* case is bad - TokenController doesn't have a constructor, it's just a small class that I want to extend, so what's the problem with classname as a parameter?

stof:
> But still don't understand why this case is bad - TokenController doesn't have a constructor

that's not true. We are injecting a dependency in it.

and the reason why parameter classes are wrong is because changing the parameter to overwrite the service assumes that your own class has **exactly** the same dependencies than the original one and the same API to inject them, even in cases where we change these dependencies in the future.
So using a parameter to provide an extension point there makes maintenance harder (or the extension point more fragile).

You should be overwriting the service definition instead.

PR #344, opened 2015-08-18
Added config to client to make possible to skip authorize step

akkonrad:
I needed this option to make process faster and automated. 
Now, when creating new client just set value:
```php
$client->setAuthorizeAutomatically(TRUE);
```
and you won't see form on oauth/v2/auth site

### DIFF
--- a/Model/Client.php
+++ b/Model/Client.php
@@ -41,6 +41,11 @@ class Client implements ClientInterface
      */
     protected $allowedGrantTypes;
 
+    /**
+     * @var boolean
+     */
+    protected $authorizeAutomatically = false;
+
     public function __construct()
     {
         $this->allowedGrantTypes = array(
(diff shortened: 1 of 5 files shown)


Model/Client.php
```diff
@@ -41,6 +41,11 @@ class Client implements ClientInterface
      */
     protected $allowedGrantTypes;
 
+    /**
+     * @var boolean
```
Soullivaneuh: `@var bool`

Model/ClientInterface.php
```diff
@@ -55,4 +55,14 @@ public function setAllowedGrantTypes(array $grantTypes);
...
      */
     public function getAllowedGrantTypes();
+
+    /**
+     * @param $value
+     */
+    public function setAuthorizeAutomatically($value);
+
+    /**
+     * @return boolean
```
Soullivaneuh: `@var bool`

Model/ClientInterface.php
```diff
@@ -55,4 +55,14 @@ public function setAllowedGrantTypes(array $grantTypes);
      * @return array
      */
     public function getAllowedGrantTypes();
+
+    /**
+     * @param $value
```
Soullivaneuh: Missing parameter type.

Issue #309, opened 2015-04-20
Show scopes in authorize.html.twig


BirknerAlex:
Hello everybody,

I want to show informations about the given scopes in the authorize_content.html.twig template.
I've did a dirty workaround in the AuthorizeController.php. This is my new return statement in line 77:

``` php
return $this->container->get('templating')->renderResponse(
    'FOSOAuthServerBundle:Authorize:authorize.html.' . $this->container->getParameter('fos_oauth_server.template.engine'),
    array(
        'form'      => $form->createView(),
        'client'    => $this->getClient(),
        'scopes'    => (!empty($form->getData()->scope)) ? explode(' ', $form->getData()->scope) : array(),
    )
);
```

I've done this because I can't find the scope informations in the client or form variable. Or is there already a way to get them inside the twig template?

Regards,
Alexander

f0086:
+1

Spomky:
+1
I will propose a PR soon to fix this issue.

BirknerAlex:
Sounds great @Spomky  :+1: 

Maybe you have a better solution then I've found. ;-)

BirknerAlex:
@Spomky Long time ago but pull request has been added for this issue. Would be great if you can review it and if it's okay merging it.

tcitworld:
This could be made easily. Anyone @Spomky @stof @GuilhemN ?

GuilhemN:
I don't this well enough to have an opinion but if we have enough positive reviews I can merge it.

PR #351, opened 2015-09-12
Added requested scopes as array to template

BirknerAlex:
Adding "scopes" variable to view template. This variable can be used to show the user which scopes has been requested by the application. This issue was discussed in #309.


Spomky:
+1
@alanbem, @stof what do you think?

BirknerAlex:
@Spomky, @alanbem, @stof 

Push

PR #357, opened 2015-09-30
Secure token path

bitgandtter:
A way to solve https://github.com/FriendsOfSymfony/oauth2-php/issues/84

### DIFF
--- a/Storage/OAuthStorage.php
+++ b/Storage/OAuthStorage.php
@@ -33,6 +33,11 @@ class OAuthStorage implements IOAuth2RefreshTokens, IOAuth2GrantUser, IOAuth2Gra
     IOAuth2GrantClient, IOAuth2GrantExtension, GrantExtensionDispatcherInterface
 {
     /**
+     * @var string
+     */
+    protected $tokenPathAuth;
+
+    /**
      * @var \FOS\OAuthServerBundle\Model\ClientManagerInterface
      */
     protected $clientManager;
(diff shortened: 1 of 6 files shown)


Storage/OAuthStorage.php
```diff
@@ -75,10 +80,11 @@ class OAuthStorage implements IOAuth2RefreshTokens, IOAuth2GrantUser, IOAuth2Gra
      * @param null|\Symfony\Component\Security\Core\User\UserProviderInterface      $userProvider
      * @param null|\Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface $encoderFactory
      */
-    public function __construct(ClientManagerInterface $clientManager, AccessTokenManagerInterface $accessTokenManager,
+    public function __construct($tokenPathAuth, ClientManagerInterface $clientManager, AccessTokenManagerInterface $accessTokenManager,
```
Spomky: There is a BC Break here as you change the signature of the constructor

Storage/OAuthStorage.php
```diff
@@ -164,12 +170,22 @@ public function checkUserCredentials(IOAuth2Client $client, $username, $password
...
+            );
+
+            if ($this->tokenPathAuth === 'plain') {
+                $encoder = $this->encoderFactory->getEncoder($user);
+
+                if ($encoder->isPasswordValid($user->getPassword(), $password, $user->getSalt())) {
+                    return $data;
+                }
+            } elseif ($this->tokenPathAuth === 'hash') {
+                if ($user->getPassword() === $password) {
```
Spomky: Using this comparison, the client is supposed to know the salted password stored in the storage.
This is not really secured: if the pasword of the user is `SJ%-qgjC]o/NWtID[u:HI#7,#Yy\d97~(DiNEPLKgyL;x?2|S` and the salted password in the storage is `8e604e3823eeec0e3209f36e7a1e5f9eeacd46b7797d1db60cc74f82aae65901` (sha256 of `SaltSJ%-qgjC]o/NWtID[u:HI#7,#Yy\d97~(DiNEPLKgyL;x?2|S`), the credential used by the client is less secured than the password.

Spomky:
I disagree with this PR as it does not solve the problem of cleartext passwords and creates a new security issue (it reduces the quality of the credentials).
IMHO, if this grant type flow does not fit on your needs, you should create a new grant type where the password is encrypted using algorithm and shared key (client side) and decrypted using the same encryption material and compared (server side).

Storage/OAuthStorage.php
bitgandtter: The parameter can be move to be the lastone and optional by default it at plain

Storage/OAuthStorage.php
bitgandtter: Its common to retreive the salt from the server with an end point to hash the plain password on the client and send it hashed to the server. The credentials indeed can be less secure but hashed there is no way a middle man can cash it with an sniffer

bitgandtter:
My intention as you say is to introduce an easy and common way to work around this plain password been sent over the air. I know this can be solve in several ways. If this solution its not the best way to do it, from your point of view; please lets try to find a better one that comes along by default and easy to configure so newest users can easily use it. Thanks

Spomky:
Look at the following gist: https://gist.github.com/Spomky/08e9b5b596ab3b9c9c3a
It has not been tested, but should help you to implement a ROPC Grant Type like, but with secured credentials transmission.

PR #358, opened 2015-10-20
Remove deprecated code and add new implementation for Symfony 3.0

ghostika:
As Symfony 3.0 will be released in a month, this bundle should be updated so it can be used with the newer version.
- `SecurityContext` removed and added `TokenStorage` for `OAuthListener`
- Test is fixed for `OAuthListener`
- In FormHandler the `bind` method was removed and `submit` was added
- The `request` service is removed, the `request` is passed in as a function argument

### DIFF
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -54,18 +54,18 @@ class AuthorizeController extends ContainerAware
 
         $event = $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::PRE_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient())
+            new OAuthEvent($user, $this->getClient($request))
         );
 
         if ($event->isAuthorizedClient()) {
-            $scope = $this->container->get('request')->get('scope', null);
+            $scope = $request->get('scope', null);
 
             return $this->container
                 ->get('fos_oauth_server.server')
                 ->finishClientAuthorization(true, $user, $request, $scope);
         }
 
-        if (true === $formHandler->process()) {
+        if (true === $formHandler->process($request)) {
             return $this->processSuccess($user, $formHandler, $request);
         }
 
@@ -73,14 +73,15 @@ class AuthorizeController extends ContainerAware
             'FOSOAuthServerBundle:Authorize:authorize.html.' . $this->container->getParameter('fos_oauth_server.template.engine'),
             array(
                 'form'      => $form->createView(),
-                'client'    => $this->getClient(),
+                'client'    => $this->getClient($request),
             )
         );
     }
 
     /**
-     * @param UserInterface        $user
+     * @param UserInterface $user
      * @param AuthorizeFormHandler $formHandler
+     * @param Request $request
      *
      * @return Response
      */
@@ -93,7 +94,7 @@ class AuthorizeController extends ContainerAware
 
         $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::POST_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient(), $formHandler->isAccepted())
+            new OAuthEvent($user, $this->getClient($request), $formHandler->isAccepted())
         );
 
         $formName = $this->container->get('fos_oauth_server.authorize.form')->getName();
@@ -124,13 +125,12 @@ class AuthorizeController extends ContainerAware
     /**
      * @return ClientInterface
      */
-    protected function getClient()
+    protected function getClient(Request $request)
     {
         if (null === $this->client) {
-            if (null === $clientId = $this->container->get('request')->get('client_id')) {
+            if (null === $clientId = $request->get('client_id')) {
                 $form = $this->container->get('fos_oauth_server.authorize.form');
-                $clientId = $this->container->get('request')
-                    ->get(sprintf('%s[client_id]', $form->getName()), null, true);
+                $clientId = $request->get(sprintf('%s[client_id]', $form->getName()), null, true);
             }
 
             $client = $this->container
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -54,7 +54,7 @@ class AuthorizeController extends ContainerAware
 
         $event = $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::PRE_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient($request))
+            new OAuthEvent($user, $this->getClient())
         );
 
         if ($event->isAuthorizedClient()) {
@@ -73,7 +73,7 @@ class AuthorizeController extends ContainerAware
             'FOSOAuthServerBundle:Authorize:authorize.html.' . $this->container->getParameter('fos_oauth_server.template.engine'),
             array(
                 'form'      => $form->createView(),
-                'client'    => $this->getClient($request),
+                'client'    => $this->getClient(),
             )
         );
     }
@@ -94,7 +94,7 @@ class AuthorizeController extends ContainerAware
 
         $this->container->get('event_dispatcher')->dispatch(
             OAuthEvent::POST_AUTHORIZATION_PROCESS,
-            new OAuthEvent($user, $this->getClient($request), $formHandler->isAccepted())
+            new OAuthEvent($user, $this->getClient(), $formHandler->isAccepted())
         );
 
         $formName = $this->container->get('fos_oauth_server.authorize.form')->getName();
@@ -125,8 +125,9 @@ class AuthorizeController extends ContainerAware
     /**
      * @return ClientInterface
      */
-    protected function getClient(Request $request)
+    protected function getClient()
     {
+        $request = $this->container->get('request_stack')->getMasterRequest();
         if (null === $this->client) {
             if (null === $clientId = $request->get('client_id')) {
                 $form = $this->container->get('fos_oauth_server.authorize.form');
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -65,7 +65,7 @@ class AuthorizeController extends ContainerAware
                 ->finishClientAuthorization(true, $user, $request, $scope);
         }
 
-        if (true === $formHandler->process($request)) {
+        if (true === $formHandler->process()) {
             return $this->processSuccess($user, $formHandler, $request);
         }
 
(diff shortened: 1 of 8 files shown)


Security/Firewall/OAuthListener.php
```diff
@@ -44,13 +44,13 @@ class OAuthListener implements ListenerInterface
...
     /**
-     * @param \Symfony\Component\Security\Core\SecurityContextInterface                      $securityContext       The security context.
-     * @param \Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface $authenticationManager The authentication manager.
-     * @param \OAuth2\OAuth2                                                                 $serverService
+     * @param \Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface   $tokenStorage          The token storage.
+     * @param \Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface        $authenticationManager The authentication manager.
+     * @param \OAuth2\OAuth2                                                                        $serverService
      */
-    public function __construct(SecurityContextInterface $securityContext, AuthenticationManagerInterface $authenticationManager, OAuth2 $serverService)
+    public function __construct(TokenStorageInterface $tokenStorage, AuthenticationManagerInterface $authenticationManager, OAuth2 $serverService)
```
stof: instead of supporting only the new API, please support both the old and the new one, so that we can keep symfony 2.3 support until the end of its LTS.

Security/Firewall/OAuthListener.php
```diff
@@ -29,9 +29,9 @@
 class OAuthListener implements ListenerInterface
 {
     /**
-     * @var \Symfony\Component\Security\Core\SecurityContextInterface
+     * @var \Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface
      */
-    protected $securityContext;
+    protected $tokenStorage;
```
stof: renaming protected properties is a BC break. I would keep the old name here for BC

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -39,15 +38,15 @@ public function isRejected()
         return !$this->form->getData()->accepted;
     }
 
-    public function process()
+    public function process(Request $request)
```
stof: Changing this signature is a BC break. What about injecting the RequestStack instead, to be able to support newer Symfony versions without breaking BC ?

Controller/AuthorizeController.php
```diff
@@ -124,13 +125,12 @@ protected function getRedirectionUrl(UserInterface $user)
     /**
      * @return ClientInterface
      */
-    protected function getClient()
+    protected function getClient(Request $request)
```
stof: changing the signature of the protected method is a BC break (and this method is meant to be an extension point). Use the RequestStack instead

ghostika:
@stof my initial idea was to create a new branch, like this bundle has also a separate branch for SF 2.0.x . That wouldn't be sufficient?

stof:
@ghostika having multiple branches is a pain: it means that either the old version is unmaintained (not good when it is the version targetting the LTS version of Symfony, as it means people using the LTS are forced to use unmaintained packages) or that we have to maintain multiple branches in parallel (which increases the maintenance work. And this is not a good idea as people maintaining the bundle can forget about this need as the bundle is maintained by the different FOS members occasionally. Maintaining multiple branches requires a strong discipline).
given that it would not be harder to keep BC and to support both APIs, it is a much better solution.
The Symfony 2.0 times were different: it was impossible to support both 2.0 and 2.1 forms at the same time (the Symfony BC policy was not yet in place), and the 2.0.x branch of the bundle effectively went unmaintained very quickly.

> If not, then the support both API should happen like that I remove the type hinting and in the service.xml in case of on-invalid I use the newer API?

you should indeed remove the typehint and replace it with instanceof checks in the constructor (as we cannot have typehints on unions).
For the injection, it is no done using ``on-invalid`` (which does not allow to inject a replacement but only to ignore it or to inject ``null``). It requires using a compiler pass to change the service definition. See FOSRestBundle which does it in its 1.7 branch.

ghostika:
@stof you are right and thanks for the example, I will change them

ghostika:
@stof done

Controller/AuthorizeController.php
```diff
@@ -79,8 +79,9 @@ public function authorizeAction(Request $request)
     }
 
     /**
-     * @param UserInterface        $user
+     * @param UserInterface $user
```
stof: please keep the alignment

Controller/AuthorizeController.php
```diff
@@ -126,11 +127,11 @@ protected function getRedirectionUrl(UserInterface $user)
      */
     protected function getClient()
     {
+        $request = $this->container->get('request_stack')->getMasterRequest();
```
stof: it should get the current request (even though it will be the same here)

DependencyInjection/Compiler/TokenStorageCompilerPass.php
```diff
@@ -0,0 +1,29 @@
...
+{
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $definition = $container->getDefinition('fos_oauth_server.security.authentication.listener');
+
+        if (interface_exists('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')) {
```
stof: you should rather check whether the service is defined

DependencyInjection/Compiler/TokenStorageCompilerPass.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
```
stof: please add the license header of the bundle

Security/Firewall/OAuthListener.php
```diff
@@ -44,11 +45,11 @@ class OAuthListener implements ListenerInterface
...
-     * @param \Symfony\Component\Security\Core\SecurityContextInterface                      $securityContext       The security context.
-     * @param \Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface $authenticationManager The authentication manager.
-     * @param \OAuth2\OAuth2                                                                 $serverService
+     * @param \Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface|\Symfony\Component\Security\Core\SecurityContextInterface   $tokenStorage          The token storage.
+     * @param \Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface        $authenticationManager The authentication manager.
+     * @param \OAuth2\OAuth2                                                                        $serverService
      */
-    public function __construct(SecurityContextInterface $securityContext, AuthenticationManagerInterface $authenticationManager, OAuth2 $serverService)
+    public function __construct($securityContext, AuthenticationManagerInterface $authenticationManager, OAuth2 $serverService)
     {
```
stof: please add the validation of the type inside the constructor to forbid wrong types (allowing us to typehint ``TokenStorageInterface`` once we drop support for Symfony 2.3)

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -23,10 +24,10 @@ class AuthorizeFormHandler
     protected $request;
     protected $form;
 
-    public function __construct(FormInterface $form, Request $request)
+    public function __construct(FormInterface $form, RequestStack $requestStack)
     {
         $this->form = $form;
-        $this->request = $request;
+        $this->request = $requestStack->getMasterRequest();
```
stof: Please don't get the request in the constructor. This is broken as your service is not request-scoped and so might be instantiated before the request is available for instance.
You should store the RequestStack in a private property instead, and retrieve the current request only in ``process()``

ghostika:
@stof ping

ghostika:
@stof can this b merged or do I need to change something ?

ghostika:
@stof ping

sekl:
@ghostika It seems you're missing a ```use Symfony\Component\DependencyInjection\Reference;``` in ```DependencyInjection/Compiler/TokenStorageCompilerPass.php```

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -41,13 +47,15 @@ public function isRejected()
...
+        $this->request = $this->requestStack->getCurrentRequest();
+
         $this->form->setData(new Authorize(
             $this->request->request->has('accepted'),
             $this->request->query->all()
         ));
 
         if ('POST' === $this->request->getMethod()) {
-            $this->form->bind($this->request);
+            $this->form->submit($this->request);
```
luispabon: Passing Request to submit is now deprecated, you need to swap this to handleRequest. 

In this case it doesn't matter, but if elsewhere you had to change submit for handleRequest and the method of Request could be PUT, you need to ensure that the form object was created with $options --> ['method', 'PUT'].

luispabon: This still needs fixing, won't work on S3.

Ener-Getick:
Could you update the ``.travis.yml`` to test symfony 2.8 and 3.0 against stable version?

Ener-Getick:
ping @stof @ghostika 
Any news about this PR? Should I create another PR?

BallisticPain:
It seems to me it's still currently waiting on a couple of updates...
 * Implement using `handleRequest`
   * An example of this can be seen at: http://symfony.com/doc/current/book/forms.html#handling-form-submissions
 * Missing a `use Symfony\Component\DependencyInjection\Reference;` in `DependencyInjection/Compiler/TokenStorageCompilerPass.php`

BallisticPain:
@ghostika Can you make those final changes so we can get @stof attention for a merge potential?

luispabon:
There are two pull requests into his branch, one mine and someone else's.
If no activity, I'll open another pr collecting all the changes tomorrow
morning :+1:

-- 



1 Neal Street London WC2H 9QL

www.rightster.com | @Rightster <https://twitter.com/rightster> | LinkedIn 
<https://www.linkedin.com/company/rightster>


Rightster Limited (incorporated in England with company number 07634543) 
has its registered office at 1 Neal Street London WC2H 9QL. Rightster 
Limited is a subsidiary of Rightster Group plc (incorporated in England and 
Wales with company number 08754680).

luispabon:
So I've opened a new PR here: https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/371

It has all of @ghostika's changes, plus extra fixes to what @BallisticPain commented above  by @sekl and myself.

Hope @ghostika is OK with this, but I can't wait for this much longer.

Ener-Getick:
Closed in favor of #374

PR #361, opened 2015-11-05
Fixed phpdoc annotations

chapay:


Spomky:
+1

Event/OAuthEvent.php
```diff
@@ -37,7 +37,9 @@ class OAuthEvent extends Event
     private $isAuthorizedClient;
 
     /**
-     * @param UserInterface $user
+     * @param UserInterface   $user
+     * @param ClientInterface $client
+     * @param Boolean         $isAuthorizedClient
```
Ener-Getick: please use ``bool`` here instead.

chapay: Done.
I just used the same CS which is already used in `OAuthEvent.php` and [ClientInterface.php](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Model/ClientInterface.php#L35) files.

Ener-Getick: I see but I prefer following the symfony standards 😉

Ener-Getick:
Thank you @chapay.

PR #364, opened 2015-11-22
Update index.md

dangaffey:
This brief example defining the publicId, as well as, using the token endpoint with a newly created client would have saved me a load of frustration.  I believe it adds value to the current documentation because at the moment only the authorize endpoint is discussed.

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -581,8 +581,7 @@ $client->setAllowedGrantTypes(array('token', 'authorization_code'));
 $clientManager->updateClient($client);
 ```
 
-Once you have created a client, you need to pass its `publicId` to the authorize endpoint. You also need 
-to specify a redirect uri as well as a response type. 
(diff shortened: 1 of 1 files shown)


Resources/doc/index.md
```diff
@@ -592,6 +591,16 @@ return $this->redirect($this->generateUrl('fos_oauth_server_authorize', array(
...
 
+To use your newly created client with the token endpoint, you can pass the `publicId`, `secret`, and the corresponding parameters based on your selected grant type.
+
+```ini
+https://www.example.com/oauth/v2/token?client_id=1_RandomId
+&client_secret=Secret
+&grant_type=client_credentials
+```
+
+It is important to note that the `publicId` that that you are providing as the client_id parameter, is a concatenation of the primary key for the client record, an underscore, and the random_id field from the client record.
```
Ener-Getick: There is a typo here ``that that``

PR #371, opened 2015-12-16
[Extended] Remove deprecated code and add new implementation for Symfony 3.0 

### LINKED ISSUE #366: Make it compatible with Symfony 3.0
Is it planned to make it compatible this Bundle with Symfony 3.0?

luispabon:
This is meant to supersede https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/358 as @ghostika seems to have gone away. I believe it now addresses all the issues raised, please code review.

@sekl, it has your patch in it, although not entirely sure why the link to your github user got lost on my cherry pick.

Closes #366

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ matrix:
     - php: 7.0
       env: SYMFONY_VERSION=3.0.*
   allow_failures:
+    - php: 7.0
     - php: hhvm
     - php: nightly
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
   - nightly
 
@@ -20,17 +21,18 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.6.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.7.*@dev
+      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.*@dev
+      env: SYMFONY_VERSION=2.8.*
+    - php: 7.0
+      env: SYMFONY_VERSION=2.8.*
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.x-dev as 2.6"
+      env: SYMFONY_VERSION=3.0.*
+    - php: 7.0
+      env: SYMFONY_VERSION=3.0.*
   allow_failures:
     - php: hhvm
     - php: nightly
-    - env: SYMFONY_VERSION=2.7.*@dev
-    - env: SYMFONY_VERSION=2.8.*@dev
-    - env: SYMFONY_VERSION="3.0.x-dev as 2.6"
 
 cache:
   directories:
@@ -38,7 +40,7 @@ cache:
     - $HOME/.composer/cache
 
 before_script:
-  - echo "extension=mongo.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`
+  - echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
   - composer selfupdate
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
   - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.3.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.5.*
-    - php: 5.6
-      env: SYMFONY_VERSION=2.6.*
-    - php: 5.6
       env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ php:
   - 5.4
   - 5.5
   - 5.6
-  - 7.0
   - hhvm
   - nightly
 
@@ -24,14 +23,9 @@ matrix:
       env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
-    - php: 7.0
-      env: SYMFONY_VERSION=2.8.*
     - php: 5.6
       env: SYMFONY_VERSION=3.0.*
-    - php: 7.0
-      env: SYMFONY_VERSION=3.0.*
   allow_failures:
-    - php: 7.0
     - php: hhvm
     - php: nightly
 
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -152,14 +152,17 @@ class AuthorizeController implements ContainerAwareInterface
                 $request = $this->container->get('request');
             }
 
-            if (null === $clientId = $request->get('client_id')) {
-                $form = $this->container->get('fos_oauth_server.authorize.form');
-                $clientId = $request->get(sprintf('%s[client_id]', $form->getName()), null, true);
-            }
+            $client = null;
+            if (null !== $request) {
+                if (null === $clientId = $request->get('client_id')) {
+                    $form = $this->container->get('fos_oauth_server.authorize.form');
+                    $clientId = $request->get(sprintf('%s[client_id]', $form->getName()), null, true);
+                }
 
-            $client = $this->container
-                ->get('fos_oauth_server.client_manager')
-                ->findClientByPublicId($clientId);
+                $client = $this->container
+                    ->get('fos_oauth_server.client_manager')
+                    ->findClientByPublicId($clientId);
+            }
 
             if (null === $client) {
                 throw new NotFoundHttpException('Client not found.');
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -13,10 +13,13 @@ namespace FOS\OAuthServerBundle\Controller;
 
 use FOS\OAuthServerBundle\Event\OAuthEvent;
 use FOS\OAuthServerBundle\Form\Handler\AuthorizeFormHandler;
-use OAuth2\OAuth2;
+use FOS\OAuthServerBundle\Model\ClientInterface;
 use OAuth2\OAuth2ServerException;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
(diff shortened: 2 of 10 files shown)


luispabon:
I think Travis is playing up, some of the containers running the tests are missing the mongo extension.

luispabon:
It's definitely a travis issue, they've borked their containers.

luispabon:
@stof any thoughts?

Controller/AuthorizeController.php
```diff
@@ -26,8 +26,10 @@
  *
  * @author Chris Jones <leeked@gmail.com>
  */
-class AuthorizeController extends ContainerAware
+class AuthorizeController
 {
+    use ContainerAwareTrait;
```
stof: not good as the bundle supports PHP 5.3. Implement the interface without using the trait

luispabon: :+1:

Controller/AuthorizeController.php
```diff
@@ -126,11 +129,11 @@ protected function getRedirectionUrl(UserInterface $user)
      */
     protected function getClient()
     {
+        $request = $this->container->get('request_stack')->getCurrentRequest();
```
stof: does not work on Symfony 2.3. Please keep a fallback on the ``request`` service when there is no RequestStack``.

and this should be done inside the ``if``

DependencyInjection/Compiler/TokenStorageCompilerPass.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $definition = $container->getDefinition('fos_oauth_server.security.authentication.listener');
+
+        if ($container->hasDefinition('security.token_storage')) {
+            $tokenStorageReference = new Reference('security.token_storage');
```
stof: I would keep this one by default in the XML file, and have the compiler pass replace only when there is no token storage service. this way, we can just remove the compiler pass when the BC is not needed anymore

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -23,10 +24,15 @@ class AuthorizeFormHandler
     protected $request;
     protected $form;
 
-    public function __construct(FormInterface $form, Request $request)
+    /**
+     * @var RequestStack
+     */
+    private $requestStack;
+
+    public function __construct(FormInterface $form, RequestStack $requestStack)
```
stof: not good for PHP 5.4 support

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -41,13 +47,15 @@ public function isRejected()
 
     public function process()
     {
+        $this->request = $this->requestStack->getCurrentRequest();
```
stof: storing the request in a property is wrong, as it is not scoped anymore

Form/Handler/AuthorizeFormHandler.php
luispabon: Do you really want to keep php 5.4 support? It's not maintained any longer.

Controller/AuthorizeController.php
luispabon: Same as per php 5.4... it's not supported any longer. Do you want to burden yourself with it?

Tests/Security/Firewall/OAuthListenerTest.php
```diff
@@ -98,4 +103,10 @@ public function testHandleResponse()
         $this->assertEquals($response, $ret);
     }
 
+    public function testWrongTypeForOAuthListener()
```
stof: this test should be removed IMO. We don't guarantee we will throw such exception in the future, as we will switch back to a typehint when dropping support for Symfony 2.3

Controller/AuthorizeController.php
stof: well, I'm quite sure that some people still run it when they run the Symfony 2.3 LTS. And it would not be good to force to run an unmaintained version of the bundle for something related to security.
And given that this bundle is low maintenance, I don't think maintaining 2 branches just to avoid 4 small lines of codes here. There is no burden here.

I think dropping support for them might be done at the same time we drop support for Symfony 2.3 (which reaches end of maintenance in 6 months)

Form/Handler/AuthorizeFormHandler.php
stof: Sorry, this is not about PHP 5.4. It is about Symfony 2.3 (which is still maintained). I messed my comment

Controller/AuthorizeController.php
luispabon: roger

Form/Handler/AuthorizeFormHandler.php
luispabon: :+1:

DependencyInjection/Compiler/TokenStorageCompilerPass.php
luispabon: I'm not sure I understand what you want me to do here however? token_storage is not available on symfony 2.3, and it will throw an exception if left on the xml file; likewise security.context is only available from symfony 2.6. Unless I have the wrong end of the stick here?

luispabon:
All concerns addressed except for the one regarding `TokenStorageCompilerPass`, please see comment above.

luispabon:
Okay, this bundle cannot support PHP7 for the time being as doctrine/mongo-odm and dependencies depend on the legacy mongo extension, instead of the new mongodb extension.

oauth2-php needs updating also to symfony 3 as it depends on httpfoundation 2.0

I'm trying to fix the latter.

luispabon:
https://github.com/FriendsOfSymfony/oauth2-php/pull/88

DependencyInjection/Compiler/TokenStorageCompilerPass.php
stof: no, it won't throw an error in 2.3 as your compiler pass will replace the argument here *before* the reference graph is validated.

My idea is to make the XML file valid already for newest version of Symfony, and then having the compiler pass replace the argument for Symfony 2.3 only. This way, dropping support for Symfony 2.3 means we can simply remove the compiler pass, without having to change the XML file again (while your current proposal requires the compiler pass to change the config for any version of Symfony, so removing it without changing the XML file would break the config)

luispabon: Copy that, I will try do this next.

stof:
We can totally support PHP 7, as there is no hard dependency on the ODM. It is an optional dependency.

Controller/AuthorizeController.php
```diff
@@ -127,10 +146,15 @@ protected function getRedirectionUrl(UserInterface $user)
     protected function getClient()
     {
         if (null === $this->client) {
-            if (null === $clientId = $this->container->get('request')->get('client_id')) {
+            try {
+                $request = $this->container->get('request_stack')->getCurrentRequest();
```
stof: you need to handle the case of ``$request`` being ``null`` here to be safe

luispabon:
Understood, but you must realise builds will fail, and I imagine tests also, on PHP7 until doctrine/mongo upgrades to the new extension.

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -41,13 +59,19 @@ public function isRejected()
 
     public function process()
     {
+        try {
+            $request = $this->container->get('request_stack')->getCurrentRequest();
```
stof: same here about handling of ``null``

stof:
@luispabon no. tests relying on mongo should simply be skipped if the mongo extension is not available

luispabon:
Fair do's, although that's out of scope for this particular PR. This is symfony 3.0 compat only. Builds will still fail due to composer not being able to resolve dependencies. I've moved php7 to allow_failures on travis.

.travis.yml
```diff
@@ -20,17 +21,19 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.6.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.7.*@dev
+      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.*@dev
+      env: SYMFONY_VERSION=2.8.*
+    - php: 7.0
+      env: SYMFONY_VERSION=2.8.*
```
stof: there is no reason to run tests for Symfony 2.8 twice, on 5.6 and 7.0. The bundle is already tested on both different versions of PHP, and Symfony itself too.

luispabon:
Also @stof, if you have any hand on the FOS oauth2-php, we need this https://github.com/FriendsOfSymfony/oauth2-php/pull/88 merged and a new release cut. Otherwise composer won't be able to resolve working dependencies when using this alongside symfony 3.

stof:
> Fair do's, although that's out of scope for this particular PR. This is symfony 3.0 compat only. Builds will still fail due to composer not being able to resolve dependencies. I've moved php7 to allow_failures on travis.

Given that running tests on PHP 7 is something you added, it **is** in the scope. Otherwise, you should not add PHP 7 at all in this PR, and do a separate PR for it (which might indeed be a good idea though)

luispabon:
Fair enough, php7 was on the original PR and I expanded travis config for it. I've removed it entirely, best left for a different PR. I've also bumped all symfony components to allow for 3.0. Will have a look at the token compiler pass when I have some time.

We're blocked by oauth2-php atm.

stof:
> We're blocked by oauth2-php atm.

The PR has been merged, even though it is not released yet

luispabon:
:+1:

.travis.yml
```diff
@@ -20,17 +20,14 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.6.*
```
stof: I suggest removing 2.5 and 2.6 here, as they reached end of maintenance. This will avoid making the Travis build too big.

luispabon: Will do. The travis config is also asking for php 5.4 and 5.5, although they're not being tested. 5.3 is tested however. Anything you'd like to do about this?

stof: PHP 5.4 and 5.5 *are* tested. They are running on builds which are not modifying the Symfony requirement

Form/Handler/AuthorizeFormHandler.php
luispabon: done on both accounts

DependencyInjection/Compiler/TokenStorageCompilerPass.php
luispabon: Done.

luispabon:
Addressed the remaining concerns. Still blocked on oauth2-php, unless we switch to dev_master.

luispabon:
I won't be near a computer for the next 10 days, please feel free to complete this as needed.

Ener-Getick:
Closed in favor of #374

PR #391, opened 2016-02-08
access token grant event listener

KonradRozner:


GabrielCW:
:+1:

Controller/TokenController.php
```diff
@@ -37,9 +55,25 @@ public function __construct(OAuth2 $server)
...
+            /* @var $accessToken \FOS\OAuthServerBundle\Entity\AccessToken */
+
+            $this->container->get('event_dispatcher')->dispatch(
+                OAuthTokenEvent::POST_ACCESS_TOKEN_GRANT,
+                new OAuthTokenEvent($accessToken)
+            );
+
+            return $response;
+        }
+        catch (OAuth2ServerException $e) {
```
Ener-Getick: No need to apply this ``try ... catch`` to the entire function.
just apply it on the first line.

Controller/TokenController.php
```diff
@@ -11,14 +11,32 @@
...
+     * @var ContainerInterface
+     */
+    protected $container;
+
+    /**
+     * Sets the container.
+     *
+     * @param ContainerInterface|null $container A ContainerInterface instance or null
+     */
+    public function setContainer(ContainerInterface $container = null)
```
Ener-Getick: please use constructor injection instead and support the case where the ``EventDispatcher`` is not set for BC.

KonradRozner: I'm not sure if that `container->has('event_dispatcher')` is enough for BC

jayrulez:
We need this also. Will this change be merged? soon?

Ener-Getick:
My comment is still not fixed. First of all, it is a bad practice to inject the container, you can inject only the event dispatcher. And secondly your changes break BC. You have to support the case where the event dispatcher is not injected.

KonradRozner:
I needed `fos_oauth_server.storage` that's why I passed the whole container as an argumment, now storage and dispatcher are injected separately

Spomky:
As the event dispatcher might not be available, it has to be ignored in this case in the XML file

KonradRozner:
@Spomky `on-invalid` attribute is that what you mean?

Spomky:
That is right. But it seems that your PR is already up to date

Controller/TokenController.php
```diff
@@ -24,11 +27,25 @@ class TokenController
...
+     */
+    protected $dispatcher;
+
+    /**
      * @param OAuth2 $server
+     * @param IOAuth2Storage $storage
+     * @param EventDispatcherInterface $dispatcher
      */
-    public function __construct(OAuth2 $server)
+    public function __construct(OAuth2 $server, IOAuth2Storage $storage, EventDispatcherInterface $dispatcher = null)
```
Ener-Getick: That's still a BC break as the class now needs the ``IOAuth2Storage``.

KonradRozner:
I have removed `IOAuth2Storage` injection, but I still need it for that case so I've used `Reflection` to get it. I know it's dirty but I don't see any other way to get it right without breaking BC. 
Any thoughts?

jayrulez:
We have been maintaining this out of tree for around 6 months now.

Any chance of it getting merged?

Sgoettschkes:
Would it be possible to also dispatch an event in case of an error? I'd need this to log each attempt to get a token no matter the result.

PR #392, opened 2016-02-13
Update index.md

rodrigolopezguerra:
Little note that discourage the use of security.context for symfony 2.6 and above , since has been deprecated


Ener-Getick:
IMO we should even replace the deprecated features in the documentation by their replacements as new installations will be based on sf 3.0.

PR #405, opened 2016-03-29
Added test for the CleanCommand.

Gadgetdude:
- Added a test for the `CleanCommand`.
- Added `symfony/console` to dev dependencies to be able to test it.
- Added `symfony/console` to suggestions to be able to use it for users.

Might it also be an idea to add `phpunit/phpunit` as a dev dependency for testing purposes?

### DIFF
--- a/Tests/Command/CleanCommandTest.php
+++ b/Tests/Command/CleanCommandTest.php
@@ -1,0 +1,168 @@
+<?php
+
+/*
+ * This file is part of the FOSOAuthServerBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\OAuthServerBundle\Tests\Command;
+
+use FOS\OAuthServerBundle\Command\CleanCommand;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class CleanCommandTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var \FOS\OAuthServerBundle\Command\CleanCommand
+     */
+    private $command;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Symfony\Component\DependencyInjection\ContainerInterface
+     */
+    private $container;
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        $command = new CleanCommand();
+
+        $application = new Application();
+        $application->add($command);
+
+        $this->container = $this->getMock('Symfony\Component\DependencyInjection\ContainerInterface');
+
+        $this->command = $application->find($command->getName());
+        $this->command->setContainer($this->container);
+    }
+
+    /**
+     * Delete expired tokens for classes that implement the TokenManagerInterface.
+     */
+    public function testItShouldRemoveExpiredTokensForTokenManagerInterfaces()
+    {
+        $expiredAccessTokens = 5;
+        $accessTokenManager = $this->getMock('FOS\OAuthServerBundle\Model\TokenManagerInterface');
+        $accessTokenManager
+            ->expects($this->once())
+            ->method('deleteExpired')
+            ->will($this->returnValue($expiredAccessTokens));
+
+        $expiredRefreshTokens = 3;
+        $refreshTokenManager = $this->getMock('FOS\OAuthServerBundle\Model\TokenManagerInterface');
+        $refreshTokenManager
+            ->expects($this->once())
+            ->method('deleteExpired')
+            ->will($this->returnValue($expiredRefreshTokens));
+
+        $expiredAuthCodes = 0;
+        $authCodeManager = $this->getMock('FOS\OAuthServerBundle\Model\TokenManagerInterface');
+        $authCodeManager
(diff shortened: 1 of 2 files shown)


Gadgetdude:
_Looking into why the build fails for `--prefer-lowest`._

Gadgetdude:
After some research I noticed that `symfony/dependency-injection` at the lowest preferred version (2.0.4) did not include an autoload for the classes. The fatal error thrown is correct in the sense that the classes for that component weren't loaded. This was solved/added in version 2.0.5 of the bundle. To minimize bc breaks I've added the constraint `"symfony/dependency-injection": "^2.0.5|~3.0"`.

```
PHP Fatal error:  Interface 'Symfony\Component\DependencyInjection\ContainerAwareInterface' not found in <user>/Karify.FOSOAuthServerBundle/vendor/symfony/framework-bundle/Symfony/Bundle/FrameworkBundle/Command/ContainerAwareCommand.php on line 24
```

Gadgetdude:
@Ener-Getick: Do you perhaps have some time to review the command test? I would greatly appreciate the feedback on the code (and workflow on the pull request).
My goal is to learn more so I can contribute better and more often to open source projects.

Tests/Command/CleanCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+            ->method('get')
+            ->will($this->returnCallback(function ($argument) use ($containerMap) {
+                return $containerMap[$argument];
+            }));
+
+        $tester = new CommandTester($this->command);
+        $tester->execute(array('command' => $this->command->getName()));
+        $display = $tester->getDisplay();
+
+        $this->assertRegExp(sprintf('\'Removed %d items from %s storage.\'', $expiredAccessTokens, 'Access token'), $display);
```
Ener-Getick: can't we use ``assertContains`` here instead ?

Tests/Command/CleanCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+            ->method('deleteExpired')
+            ->will($this->returnValue($expiredAuthCodes));
+
+        $containerMap = array(
+            'fos_oauth_server.access_token_manager' => $accessTokenManager,
+            'fos_oauth_server.refresh_token_manager' => $refreshTokenManager,
+            'fos_oauth_server.auth_code_manager' => $authCodeManager,
+        );
+
+        $this->container
```
Ener-Getick: What about using a real instance of the container ?

Tests/Command/CleanCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $this->command->setContainer($this->container);
+    }
+
+    /**
+     * Delete expired tokens for classes that implement the TokenManagerInterface.
+     */
+    public function testItShouldRemoveExpiredTokensForTokenManagerInterfaces()
+    {
+        $expiredAccessTokens = 5;
+        $accessTokenManager = $this->getMock('FOS\OAuthServerBundle\Model\TokenManagerInterface');
```
Ener-Getick: I would prefer to use a ``@dataProvider`` to avoid duplication.
```php
/**
 * @dataProvider classProvider
 */
public function testMyFeature($class) {
    $this->getMock($class);
    // ...
}

public function classProvider()
{
    return array(
        array('FOS\OAuthServerBundle\Model\TokenManagerInterface'),
        array('FOS\OAuthServerBundle\Model\AuthCodeManagerInterface'),
    );
}
```

Ener-Getick:
Otherwise looks good to me ! Thanks for this contribution :smile:

Tests/Command/CleanCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+namespace FOS\OAuthServerBundle\Tests\Command;
+
+use FOS\OAuthServerBundle\Command\CleanCommand;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class CleanCommandTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var \FOS\OAuthServerBundle\Command\CleanCommand
```
Ener-Getick:
```php
/**
 * @var CleanCommand
 */
```
would be enough here

Tests/Command/CleanCommandTest.php
Gadgetdude: That is something I haven't thought of. For me a "unit" is just one class (and sometimes additional classes that are needed to run the tests), but since the container is basically a value object, it should be fine to use the real class. I will incorporate the changes in the next commit.

Tests/Command/CleanCommandTest.php
Gadgetdude: That would be much cleaner for these matches indeed. It was leftover code from the `assertNotRegExp` from the last test where I actually needed to use `\d+` for matching.

Tests/Command/CleanCommandTest.php
Gadgetdude: Fixed in next commit.

Tests/Command/CleanCommandTest.php
Gadgetdude: Thanks for the tip. I've refactored it. The test class in now much shorter and more readable.

Gadgetdude:
@Ener-Getick: Thanks for the review! I've updated the branch accordingly.

Ener-Getick:
Thank you @Gadgetdude :-)

PR #410, opened 2016-06-03
Added support for different user classes

n0rbyt3:
In my project I have two different user classes implementing `UserInterface`: `User` and `Account`. An user is authenticated by username and password and an account is authenticated by an unique token. The user provider supports both classes.

`AccessToken`, `RefreshToken` and `AuthCode` are extended by an additional field "account" which is mapped to the account. But the `OAuthProvider` always uses `AccessToken::getUser()` to determine the user. The method cannot get overriden, otherwise Doctrine will fail writing the entities. Also this method is not part of `IOAuth2AccessTokenInterface`.

But `IOAuth2AccessToken` provides the `getData()` method which returns the user by default. The method is also safe to override. Here you can check which entity relates to the token.


dinamic:

PR #413, opened 2016-06-13
Fix undefined variable

phoenixgao:


dinamic:
I am afraid you got it wrong.

The `$strong` variable comes from this line:

```
$bytes = openssl_random_pseudo_bytes(32, $strong);
```

As mentioned in the php.net documentation for [openssl_random_pseudo_bytes()](http://php.net/openssl_random_pseudo_bytes):

> If passed into the function, this will hold a boolean value that determines if the algorithm used was "cryptographically strong", e.g., safe for usage with GPG, passwords, etc. TRUE if it did, otherwise FALSE

dinamic:
How did you came to conclusion that the variable is undefined?

dkarlovi:
Obsolete.

Issue #431, opened 2016-09-15
Move template path to config.yml


eretica:
Hey. I'm afraid my expressions may be rude or hard to read, because I'm not so good at English. But please, please be patient

I make the proposal. 
that is render twig path move to config.yml

herewith Freely setting view file.

https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Controller/AuthorizeController.php
Line: 90
<pre><code>
// BREFORE
return $this->container->get('templating')->renderResponse(
        'FOSOAuthServerBundle:Authorize:authorize.html.'.$this->container->getParameter('fos_oauth_server.template.engine'),
        array(
            'form'   => $form->createView(),
            'client' => $this->getClient(),
        )
    );

// AFTER
return $this->container->get('templating')->renderResponse(
        '$this-container->getParameter('fos_oauth_server.template.authorize').$this->container->getParameter('fos_oauth_server.template.engine'),
        array(
            'form'   => $form->createView(),
            'client' => $this->getClient(),
        )
    );
</code></pre>

I so creating pull request..

PR #432, opened 2016-09-15
Move authorize template path move to config file

eretica:
#431

I tried writing


eretica:
Allowed memory size of *** error...

what should I do..

PR #451, opened 2016-12-22
Override controller bug was fixed

t3ran13:
Override controller bug was fixed

### DIFF
--- a/Controller/TokenController.php
+++ b/Controller/TokenController.php
@@ -12,25 +12,13 @@
 namespace FOS\OAuthServerBundle\Controller;
 
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use OAuth2\OAuth2;
 use OAuth2\OAuth2ServerException;
 use Symfony\Component\HttpFoundation\Response;
 
-class TokenController
+class TokenController extends Controller
 {
-    /**
-     * @var OAuth2
-     */
-    protected $server;
-
-    /**
-     * @param OAuth2 $server
-     */
-    public function __construct(OAuth2 $server)
-    {
-        $this->server = $server;
-    }
-
     /**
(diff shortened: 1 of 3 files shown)


Spomky:
What are the benefits for such a PR?
From my POV, it tightly couples the token endpoint with the controller provided by the Symfony Framework i.e. now a lot of unnecessary methods are available.
DI should be kept instead. Main reasons explained [in this serie of blog posts](http://php-and-symfony.matthiasnoback.nl/2014/06/how-to-create-framework-independent-controllers/).


@Ener-Getick WDYT?

t3ran13:
Is it FOS or not???
http://symfony.com/doc/current/bundles/inheritance.html
https://symfony.com/doc/current/controller/service.html

"Defining controllers as services is not officially recommended by Symfony"

t3ran13:
In your own app you can do what you want, but it is not acceptable for FOS

GuilhemN:
@Spomky i missed the notification sorry... I agree with you, I prefer keeping the controller as a service.

@t3ran13 a lot of things has been done lately making even easier to use controllers as services, it is clearly supported by symfony. And this recommandation is mostly for newcomers anyway.

What is the bug you're talking about ? If this fixes a bug maybe we need to make something lazy but i don't think this is the right solution.

t3ran13:
it have to be able to override Token controller. When you use service you block this possibility, it is not acceptable for FOS bundle. No other ways exist to override controller. 

I use controllers as a services in my own bundles too, but it is FOS bundle and it have to be in Symfony architecture style and implements Symfony standards

GuilhemN:
@t3ran13 just decorate its service if you want to extend it.

Spomky:
Since sf 3.3, the recommendation mentionned above has been removed and controllers are already registered as services.

t3ran13:
Thx

PR #454, opened 2017-02-09
Pass $user to authorize view

Alsciende:
By passing $user to the authorize view, it will be possible to personalize the oauth challenge with some user information.


Spomky:
What about `{{ app.user }}`?

Alsciende:
Oh wow, sure, I totally forgot about that. Thanks!

PR #462, opened 2017-04-24
Allow the configure without authorize

TNAJanssen:
When you have authorize false don't try to load the form!


dinamic:

Issue #465, opened 2017-05-18
Impossible to disable authorize


vincentchalamon:
According to [configuration](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/DependencyInjection/Configuration.php#L65), it should be possible to disable `authorize` key. But if I do, I get the following error:
> You have requested a non-existent service "fos_oauth_server.authorize.form"

I'm trying to migrate to Symfony 3.3, which disables forms by default, and I don't want forms & authorize neither (it's a private API).

TNAJanssen:
https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/462

florianajir:
can this is be fixed ? I don't want to require dependencies that I don't use... (twig, templating, etc.)

elvismdev:
Trying to install the bundle in SF4 but got this error, I think it's related. Prefer to not install extra packages I don't need like twig.

```
The service "fos_oauth_server.authorize.form" has a dependency on a non-existent service "form.factory".
```

vincentchalamon:
Fixed in https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/554

PR #466, opened 2017-05-18
[Bug fix] Allow to disable authorize

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #465
| License       | MIT

### DIFF
--- a/DependencyInjection/FOSOAuthServerExtension.php
+++ b/DependencyInjection/FOSOAuthServerExtension.php
@@ -85,9 +85,11 @@ class FOSOAuthServerExtension extends Extension
             $ormEntityManagerDefinition->setFactory([new Reference('doctrine'), 'getManager']);
         }
 
-        if (!empty($config['authorize'])) {
-            $this->loadAuthorize($config['authorize'], $container, $loader);
+        if (empty($config['authorize'])) {
+            return;
         }
+
+        $this->loadAuthorize($config['authorize'], $container, $loader);
 
         // Authorize form factory definition
         // TODO: Go back to xml configuration when bumping the requirement to Symfony >=2.6


vincentchalamon:
Bug on Travis PHPUnit:
> PHP Fatal error:  Class 'PHPUnit\Framework\TestCase' not found in /home/travis/build/FriendsOfSymfony/FOSOAuthServerBundle/vendor/symfony/symfony/src/Symfony/Component/Form/Test/FormIntegrationTestCase.php on line 22

DependencyInjection/FOSOAuthServerExtension.php
```diff
@@ -88,10 +88,12 @@ public function load(array $configs, ContainerBuilder $container)
             }
         }
 
-        if (!empty($config['authorize'])) {
-            $this->loadAuthorize($config['authorize'], $container, $loader);
+        if (empty($config['authorize'])) {
+            return;
         }
 
+        $this->loadAuthorize($config['authorize'], $container, $loader);
```
dinamic: Isn't this change doing the exact same thing?

vincentchalamon: It's not the same thing as my version exit the function, so lines 99->104 won't be executed. In original version, they are executed even if authorize feature is disabled, which is a bug IMO.
Still waiting for the team to review & merge this PR

dinamic: Version 1.6 is about to be tagged soon. We may be able to squeeze this in.

I think you would have better chance of this getting thru if you solve the conflicts. Unit testing the change would be nice too.

dinamic:
This seems to duplicate #462, but it's solution relies on returning early and I fancy that. Once this get's merged in, we should be able to close #462 and #465.

DependencyInjection/FOSOAuthServerExtension.php
vincentchalamon: Conflicts solved, but this class is not unit tested

dinamic: You are right. I was mislead by the presence of the test file.

dinamic:
@dkarlovi, could we have this in the milestone for 1.6?

florianajir:
please merge it!

dkarlovi:
Fixed in #554.

PR #475, opened 2017-08-07
Fixed Propel undefined use

qkdreyer:
That was causing an internal server error in a case where I was using another Client class without the correct or missing use.


qkdreyer:
I've just noticed that all Propel files are afflicted by the same "issue", if it is indeed a real one.

PR #482, opened 2017-11-12
added cascade on delete

iCodr8:
When a user will be deleted, then the tokens should be deleted too. Otherwise it is not possible to remove a user, when a token exists.


dinamic:

@dkarlovi, how about this feature? Maybe we can include it in 2.0 as it breaks behaviour.

iCodr8:
Rebase is done!

dkarlovi:
@iCodr8 thanks!

PR #485, opened 2017-11-18
Adding Symfony4 support

dnahrebecki:
Hi guys, in this PR I'm adding support for Symfony4. The "biggest" change was to change `AuthorizeController` to do not rely on container anymore, cause anyway most of services are not accessible directly. The idea was to move it to a controller as a service approach.
Also:

- dropped check for symfony 2.3 and 2.7 in travis as you guys have a check for 2.8 (consulted with @weaverryan)
- RequestStack was introduced in symfony 2.4 so we can drop checks for BC
- removed deprecations in phpunit (`getMock()`)
- added workaround for mongodb extension (consulted with @Simperfit)

However, i didn't test it on live app.

#SymfonyConHackday2017


.travis.yml
```diff
@@ -3,18 +3,22 @@ language: php
 sudo: false
 
 php:
-  - 5.3
```
dnahrebecki: needed to remove it due to https://docs.travis-ci.com/user/reference/trusty#PHP-images

composer.json
```diff
@@ -42,6 +44,9 @@
         "symfony/form" : "Needed to be able to use the AuthorizeFormType",
         "symfony/console": "Needed to be able to use commands"
     },
+    "provide": {
+        "ext-mongo": "^1.6.14"
```
dnahrebecki: see https://github.com/api-platform/core/pull/1293/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

dnahrebecki:
changes related to dropping symfony 2.3 and 2.7 discussed with @weaverryan

dnahrebecki:
changes related to mongodb extension support discussed @simperfit

composer.json
alcaeus: If you're providing `ext-mongo`, you also have to install a package that provides the API (see https://github.com/api-platform/core/pull/1293/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R28).

alcaeus: To follow up on our conversation at the hack day, it is smarter to not install require the adapter in general, but rather on demand in travis.yml. One example can be found in the MongoDB ODM repository: https://github.com/doctrine/mongodb-odm/blob/cc122d144f32d661be853bf5af0a44acd63602a7/.travis.yml#L42..L44

cored0wn:
Great work! :heart_eyes: I'm eagerly awaiting the implementation as I'm sitting at a fresh Symfony4 project.

composer.json
dnahrebecki: Thanks, I'll apply changes within next few days. Btw. according to our discussion in Cluj, have you had a chance to test my changes against your app and testsuites?

dnahrebecki:
guys, looks like we have green build. @alcaeus would be great if you could test it against your app/testsuites

composer.json
alcaeus: I've hit a brick wall trying to update to Symfony 3.4. I'll run it against 3.3 to be sure ;)

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -51,16 +51,6 @@ public function __construct(FormInterface $form, $requestStack = null)
         $this->requestStack = $requestStack;
     }
 
-    /**
-     * Sets the container.
-     *
-     * @param ContainerInterface|null $container A ContainerInterface instance or null
-     */
-    public function setContainer(ContainerInterface $container = null)
```
alcaeus: Note: this is a BC break - if you want to have the changes in a patch release, you should instead deprecate the method.

Resources/config/authorize.xml
```diff
@@ -20,7 +20,21 @@
 
         <service id="fos_oauth_server.authorize.form.handler.default" class="FOS\OAuthServerBundle\Form\Handler\AuthorizeFormHandler">
             <argument type="service" id="fos_oauth_server.authorize.form" />
-            <argument type="service" id="request_stack" on-invalid="ignore" />
+            <argument type="service" id="request_stack" />
```
alcaeus: Since you're already breaking BC by removing `setContainer`, you should also change the signature of `AuthorizeFormHandler::__construct`. The other option is keeping this optional and triggering a deprecation notice that the argument will be required soon.

composer.json
```diff
@@ -18,23 +18,25 @@
     "require": {
         "php": "^5.3.3|^7.0",
```
alcaeus: You're dropping the 5.3 build in .travis.yml, so you should also drop support for 5.3 here. I would also suggest dropping PHP 5.x completely, but that depends on what the maintainers here want to maintain.

Form/Handler/AuthorizeFormHandler.php
dkarlovi: Agreed, we can't break BC here.

Controller/AuthorizeController.php
```diff
@@ -28,66 +35,147 @@
...
+     * @param AuthorizeFormHandler   $authorizeFormHandler
+     * @param OAuth2                 $oAuth2Server
+     * @param EngineInterface        $templating
+     * @param TokenStorageInterface  $tokenStorage
+     * @param Router                 $router
+     * @param ClientManagerInterface $clientManager
+     * @param EventDispatcher        $eventDispatcher
+     * @param string                 $templateEngineType
      */
-    public function setContainer(ContainerInterface $container = null)
```
dkarlovi: @dnahrebecki as commented earlier, please ensure you do NOT break BC in this PR, I'm planning on starting a 2.x branch rather soon and revamping this lib.

alcaeus:
@dkarlovi are you taking over maintenance of this bundle? If so, thank you very much! 🍻

dkarlovi:
@alcaeus I'll do my best, help is super-appreciated. :1st_place_medal: I see this lib as quite valuable and am sad it gets ignored currently.

Roughly:
* PHP 7.x, probably PHP 7.2+
* bring it up to speed with modern standards: run PHPStan, CS Fixer and stuff on it from CI
* update to latest versions of all deps
* cleanup PRs and issues

EneasVVA:
Check OAuthFactory Class and use ChildDefinition instead DefinitionDecorator because is deprecated in SF 4.0

dkarlovi:
Expanded upon and merged in #497

SebScoFr:
@dkarlovi, @EneasVVA is right, https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/DependencyInjection/Security/Factory/OAuthFactory.php#L34 still makes reference to DefinitionDecorator on master, which brakes my SF4 setup.
```
Fatal error: Uncaught Symfony\Component\Debug\Exception\ClassNotFoundException: Attempted to load class "DefinitionDecorator" from namespace "Symfony\Component\DependencyInjection".
Did you forget a "use" statement for another namespace? in /var/www/graphql-api/vendor/friendsofsymfony/oauth-server-bundle/DependencyInjection/Security/Factory/OAuthFactory.php:34
```

dkarlovi:
Yes, it seems tests leave plenty to be desired here. PHPStan caught this in my #499 so it'll be fixed, but we need it fixed here too, only in a BC way.

You can open a PR against `1.5` and fix it, use the class which is available to make it work on all Symfony versions supported.

SebScoFr:
Well I'm using master at the moment. When do you plan to merge #499?
Many thanks

dkarlovi:
Master will be 2.x where I plan to do a major overhaul (as seen there). This is WIP so will not be merged for some time.

`1.5` will be the continuation of the current codebase and the only feature which I think it gets is Symfony 4 support, branch 2.x will (ironically) drop Symfony 2.x support. I'd like to get S4 support ASAP and release.

SebScoFr:
So `1.5` supports `"symfony/dependency-injection": "^2.8|~3.0|^4.0"`
However `ChildDefinition` is only mandatory on `4.0`, and doesn't exist on `2.8` for example.
How do you plan to fix https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/DependencyInjection/Security/Factory/OAuthFactory.php on `1.5` so that it can support all versions?

dkarlovi:
Compatiblity is being worked on in #502

PR #497, opened 2018-01-03
Symfony 4.x compatibility

wackyMole:
This PR is a modified version of #485 that replaces the two `setContainer()` methods and addresses a few of the other comments.

Unfortunately, I wasn't able to test on 4.0 yet. Will update if I get to it before someone else does.

Please let me know if I need to make further changes.


.travis.yml
```diff
@@ -3,26 +3,27 @@ language: php
...
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
   - hhvm
   - 7.0
+  - 7.1
+  - 7.2
```
wackyMole: Added 7.2

.travis.yml
```diff
@@ -3,26 +3,27 @@ language: php
...
-      env: SYMFONY_VERSION=2.3.*
-    - php: 5.6
-      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
     - php: 5.6
       env: SYMFONY_VERSION=3.0.* COVERAGE=true
+    # test the latest release (including beta releases)
+    - php: 7.2
+      env: DEPENDENCIES=beta
```
wackyMole: Latest beta release is tested on 7.2

Controller/AuthorizeController.php
```diff
@@ -50,44 +114,85 @@ public function setContainer(ContainerInterface $container = null)
         $this->container = $container;
```
wackyMole: Put the `setContainer()` method back.

Not sure if I went far enough though. Now `setContainer()` sets the container, but I didn't revert the dependency injection changes. So the container can be set, but it's never used.

.travis.yml
```diff
@@ -32,8 +33,11 @@ cache:
     - $HOME/.composer/cache/files
 
 before_script:
-  - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; fi;
-  - echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - if [ "$DEPENDENCIES" = "beta" ]; then composer config minimum-stability beta; fi;
+  - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; echo "memory_limit=2G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi;
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "5." ]]; then yes '' | pecl -q install -f mongo-stable; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then pecl install -f mongodb-stable; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then composer config "platform.ext-mongo" "1.6.16" && composer require alcaeus/mongo-php-adapter; fi
```
wackyMole: Not sure if this is what @alcaeus meant here.

>This was changed recently, but you shouldn't use ignore-platform-reqs but instead set a platform config: composer config "platform.ext-mongo" "1.6.16" && composer require alcaeus/mongo-php-adapter

Resources/config/authorize.xml
```diff
@@ -20,7 +20,21 @@
 
         <service id="fos_oauth_server.authorize.form.handler.default" class="FOS\OAuthServerBundle\Form\Handler\AuthorizeFormHandler">
             <argument type="service" id="fos_oauth_server.authorize.form" />
-            <argument type="service" id="request_stack" on-invalid="ignore" />
+            <argument type="service" id="request_stack" />
```
wackyMole: I'm not exactly sure what @alcaeus meant here. Should the removal of `on-invalid="ignore"` be reverted?

composer.json
```diff
@@ -16,23 +16,24 @@
         }
     ],
     "require": {
-        "php": "^5.3.3|^7.0",
+        "php": "^5.4|^7.0",
```
wackyMole: Drop support for 5.3.3 since the build for 5.3 was removed. I didn't drop support for 5.x entirely since I don't know the desires of the maintainers.

composer.json
```diff
@@ -16,23 +16,24 @@
...
         "propel/propel1": "^1.6.5",
         "phing/phing": "~2.4",
         "doctrine/mongodb-odm": "~1.0",
         "doctrine/doctrine-bundle": "~1.0",
-        "doctrine/orm": "~2.2"
+        "doctrine/orm": "~2.2",
+        "phpunit/phpunit": "~4.8|~5.0"
     },
     "suggest": {
         "doctrine/doctrine-bundle": "*",
```
wackyMole: Dropped the suggestion for `ext-mongodb`

composer.json
dkarlovi: This is fine for now, the new branch will switch to 7.1+

dkarlovi:
Thanks @dnahrebecki @wackyMole

Controller/AuthorizeController.php
wackyMole: Before these changes, things like the request, session, `AuthorizeFormHandler`, etc. were retrieved from the container as needed. Now those things are injected into the constructor method and so the container isn't actually used in this class.

dkarlovi:
BTW it seems (at least on Travis) this is currently uninstallable because https://github.com/FriendsOfSymfony/oauth2-php/pull/103 is still pending.

.travis.yml
alcaeus: This is exactly what I meant, good job 👍

PR #504, opened 2018-01-08
Fixes to the security doc to make it compatible with sf3

dinamic:
Symfony 2.6 deprecates the SecurityContext and Symfony 3.x eliminates it, rendering the sample code useless. I have adapted the example and applied the necessary fixes.


dkarlovi:
Thank you @dinamic

PR #507, opened 2018-01-10
[WIP] Improved unit tests

dinamic:
As mentioned in #505, we need better unit tests so we can upgrade from one version of Symfony to the other with confidence. I will try to make things better one piece at a time, currently focussed on improving the unit tests. Some pieces are not really testable, so small scale refactoring would also occur.

This is a work in progress pull request.

I am creating it so people would know that I am working on the unit tests and we won't be working on the same thing at the same time.

PR #510, opened 2018-01-12
PHPCS: Adding dangling semicolon for multi-line chained statements

dinamic:
As discussed in #507.

I have reviewed the changes and they seem reasonable.


dkarlovi:
Thank you @dinamic

PR #512, opened 2018-01-12
Unit tests for 1.6

dinamic:
Backported unit tests as discussed in #507.


.travis.yml
```diff
@@ -3,9 +3,6 @@ language: php
 sudo: false
 
 php:
-  - 5.4
```
dkarlovi: Why? We currently have support for 5.x in 1.6, don't want to remove it if not required.

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -71,26 +71,34 @@ public function isRejected()
...
+            return false;
+        }
+
+        $this->form->setData(new Authorize(
+            $request->request->has('accepted'),
+            $request->query->all()
+        ));
+
+        if ('POST' !== $request->getMethod()) {
+            return;
```
dkarlovi: Why not false here?

Tests/Document/TokenManagerTest.php
```diff
@@ -11,62 +11,178 @@
...
-    protected $dm;
+    /**
+     * @var string
+     */
+    protected $className;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|DocumentManager
+     */
+    protected $documentManager;
```
dkarlovi: Nice, proper names. :+1:

.travis.yml
dkarlovi: Besides this, LGTM.

dinamic: You are correct. This will get fixed.

dkarlovi:
Once this goes green, we can merge. :+1:

Form/Handler/AuthorizeFormHandler.php
dinamic: I will be fixing this one. Dunno how I missed it.

dinamic:
Oh, wow.. I never expected 5.4 to be so bad ass.. I feel like I've been using statements like 'RequestStack::class' to get the classname for ages, but seems like this has been supported since 5.5.

I have been using this all around the place and am not sure I will be willing to spend all the required time to manually process the lines.

Would it sound reasonable to bump the requirement a bit and drop support for PHP 5.4 in 1.6? Maybe not.

dkarlovi:
Sure, it's reasonable a reason to drop support. People using 5.4 will just not get 1.6 as an option. :+1:

dkarlovi:
Just remembered, CS Fixer can do it for you. :)

dkarlovi:
Replace ::class with a string, that is.

dinamic:
@dkarlovi, i was actually looking for the option in CS fixer and was going to revert the change in travis. I would be happy to bump the requirements tho.

Do you think its reasonable to bump them? Functionality wise 1.6 and 1.5 would be identical, so people stuck on 1.5 should have no issues with that. Right?

dkarlovi:
I think 5.5 is reasonable, can't expect miracles here.

For the Definition Decorator we already have a workaround, can we do it in tests too?

dkarlovi:
Can't target another branch for Symfony v4 as 1.6 is specifically created to allow Symfony 4 in 1.x

.travis.yml
dkarlovi: Note, also in composer.json

FOSOAuthServerBundle.php
```diff
@@ -16,22 +16,31 @@
...
+        $this->kernelVersion = Kernel::VERSION;
         $this->extension = new FOSOAuthServerExtension();
     }
 
     public function build(ContainerBuilder $container)
     {
         parent::build($container);
 
-        if (version_compare(Kernel::VERSION, '2.1', '>=')) {
+        if (version_compare($this->kernelVersion, '2.1', '>=')) {
```
dkarlovi: Why this indirection?

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -71,26 +71,34 @@ public function isRejected()
...
+        ));
+
+        if ('POST' !== $request->getMethod()) {
+            return false;
         }
 
-        return false;
+        $this->form->handleRequest($request);
+        if (!$this->form->isValid()) {
+            return;
```
dkarlovi: Same here, false? If both are fixed, we can omit the void in `@return`

FOSOAuthServerBundle.php
dinamic: I needed a way to be able to test behaviour with different versions. Here's the test:

https://github.com/dinamic/FOSOAuthServerBundle/blob/feature/unit-tests-1.6/Tests/FOSOAuthServerBundleTest.php#L37

I may have named the test the wrong way, but Kernel version is indication of the framework version being used, right?

I had that presumption when naming the tests and was thinking that I can even delete this if statement, but wasn't sure whether it won't break functionality.

dinamic:
Green build! :)

Form/Handler/AuthorizeFormHandler.php
```diff
@@ -126,14 +134,14 @@ protected function onSuccess()
...
     private function getCurrentRequest()
     {
-        if (null !== $this->requestStack) {
-            if ($this->requestStack instanceof Request) {
-                return $this->requestStack;
-            } else {
-                return $this->requestStack->getCurrentRequest();
-            }
+        if (null === $this->requestStack) {
+            return $this->container->get('request');
```
dkarlovi: Not sure if this service is public in Symfony 4.

dkarlovi:
Thank you @dinamic

dkarlovi:
@dinamic can you create a PR to merge current 1.6 into master? There are some conflicts created with this PR so it's not a simple "merge branch" situation, unfortunately.

dinamic:
Sure, I just created #517

PR #516, opened 2018-01-15
Raise PHPStan's level from 2 to 3

dkarlovi:
See #513.

### DIFF
--- a/Document/AuthCodeManager.php
+++ b/Document/AuthCodeManager.php
@@ -37,8 +37,12 @@ class AuthCodeManager extends BaseAuthCodeManager
 
     public function __construct(DocumentManager $dm, $class)
     {
+        // NOTE: bug in Doctrine, hinting DocumentRepository|ObjectRepository when only DocumentRepository is expected
+        /** @var DocumentRepository $repository */
+        $repository = $dm->getRepository($class);
+
         $this->dm = $dm;
-        $this->repository = $dm->getRepository($class);
+        $this->repository = $repository;
         $this->class = $class;
     }
 
--- a/Document/ClientManager.php
+++ b/Document/ClientManager.php
@@ -37,8 +37,12 @@ class ClientManager extends BaseClientManager
 
     public function __construct(DocumentManager $dm, $class)
     {
+        // NOTE: bug in Doctrine, hinting DocumentRepository|ObjectRepository when only DocumentRepository is expected
+        /** @var DocumentRepository $repository */
+        $repository = $dm->getRepository($class);
+
         $this->dm = $dm;
-        $this->repository = $dm->getRepository($class);
+        $this->repository = $repository;
         $this->class = $class;
     }
 
--- a/Document/TokenManager.php
+++ b/Document/TokenManager.php
@@ -37,8 +37,12 @@ class TokenManager extends BaseTokenManager
 
     public function __construct(DocumentManager $dm, $class)
     {
+        // NOTE: bug in Doctrine, hinting DocumentRepository|ObjectRepository when only DocumentRepository is expected
+        /** @var DocumentRepository $repository */
+        $repository = $dm->getRepository($class);
+
         $this->dm = $dm;
-        $this->repository = $dm->getRepository($class);
+        $this->repository = $repository;
         $this->class = $class;
     }
 
--- a/Entity/ClientManager.php
+++ b/Entity/ClientManager.php
@@ -13,8 +13,7 @@ declare(strict_types=1);
 
 namespace FOS\OAuthServerBundle\Entity;
 
-use Doctrine\Common\Persistence\ObjectManager;
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\EntityRepository;
 use FOS\OAuthServerBundle\Model\ClientInterface;
 use FOS\OAuthServerBundle\Model\ClientManager as BaseClientManager;
@@ -22,7 +21,7 @@ use FOS\OAuthServerBundle\Model\ClientManager as BaseClientManager;
 class ClientManager extends BaseClientManager
 {
     /**
-     * @var EntityManager
+     * @var EntityManagerInterface
      */
     protected $em;
 
@@ -36,10 +35,14 @@ class ClientManager extends BaseClientManager
      */
     protected $class;
 
-    public function __construct(ObjectManager $em, $class)
+    public function __construct(EntityManagerInterface $em, $class)
     {
+        // NOTE: bug in Doctrine, hinting EntityRepository|ObjectRepository when only EntityRepository is expected
+        /** @var EntityRepository $repository */
+        $repository = $em->getRepository($class);
+
         $this->em = $em;
-        $this->repository = $em->getRepository($class);
+        $this->repository = $repository;
         $this->class = $class;
     }
 
--- a/Entity/TokenManager.php
+++ b/Entity/TokenManager.php
@@ -13,8 +13,7 @@ declare(strict_types=1);
 
 namespace FOS\OAuthServerBundle\Entity;
 
-use Doctrine\Common\Persistence\ObjectManager;
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\EntityRepository;
 use FOS\OAuthServerBundle\Model\TokenInterface;
 use FOS\OAuthServerBundle\Model\TokenManager as BaseTokenManager;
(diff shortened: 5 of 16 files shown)


Document/AuthCodeManager.php
```diff
@@ -23,31 +24,31 @@ class AuthCodeManager extends BaseAuthCodeManager
     /**
      * @var DocumentManager
      */
-    protected $dm;
+    protected $documentManager;
 
     /**
-     * @var DocumentRepository
+     * @var ObjectManager|DocumentRepository
```
dinamic: This seems false.

Document/ClientManager.php
```diff
@@ -23,31 +24,31 @@ class ClientManager extends BaseClientManager
     /**
      * @var DocumentManager
      */
-    protected $dm;
+    protected $documentManager;
 
     /**
-     * @var DocumentRepository
+     * @var ObjectManager|DocumentRepository
```
dinamic: This one needs fixing IMO. Should be `DocumentRepository` only.

Document/TokenManager.php
```diff
@@ -23,31 +24,31 @@ class TokenManager extends BaseTokenManager
     /**
      * @var DocumentManager
      */
-    protected $dm;
+    protected $documentManager;
 
     /**
-     * @var DocumentRepository
+     * @var ObjectManager|DocumentRepository
```
dinamic: This seem recurring. Should be `DocumentRepository` IMO.

Entity/ClientManager.php
```diff
@@ -24,31 +24,31 @@ class ClientManager extends BaseClientManager
     /**
      * @var EntityManager
      */
-    protected $em;
+    protected $entityManager;
 
     /**
-     * @var EntityRepository
+     * @var ObjectManager|EntityRepository
```
dinamic: This should be set to `EntityRepository` only.

Entity/TokenManager.php
```diff
@@ -24,31 +24,31 @@ class TokenManager extends BaseTokenManager
     /**
      * @var EntityManager
      */
-    protected $em;
+    protected $entityManager;
 
     /**
-     * @var EntityRepository
+     * @var ObjectManager|EntityRepository
```
dinamic: Again, `EntityRepository` only.

Entity/ClientManager.php
dkarlovi: Problem is, `EntityManager::getRepository` is [documented as returning both](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/EntityManager.php#L705).


Asked @ocramius on Twitter, [he thought it's different](https://twitter.com/Ocramius/status/951845101998833664). Changed at https://github.com/doctrine/doctrine2/commit/99db207a9f1214ad483efd81341279a550526358, don't know why.

Ocramius: This should be `EntityRepository`. `EntityRepository` implements `ObjectRepository`, so there is no need to change anything here, as the type is compatible.

dkarlovi: I think the issue is, you're not guaranteed to get an `EntityRepository` from `EntityManager::getRepository` but an `ObjectRepository`.

That's the reason PHPStan is complaining, might be wrong though. @ondrejmirtes, mind confirming/denying this?

dinamic: Still, `ObjectRepository` is not `ObjectManager`. :)

dkarlovi: :see_no_evil: now I see what you mean. :+1: That's clearly an error, how on earth did it pass the build.

It's supposed to be `ObjectRepository`.

JanTvrdik: Note that `ObjectRepository|EntityRepository` is the same as just saying `ObjectRepository`.

ondrejmirtes: @dkarlovi What exactly is PHPStan complaining about?

dkarlovi: ObjectRepository cannot be assigned to EntityRepository.

ondrejmirtes: Yes, makes sense, so `ObjectRepository` should be used everywhere.

dkarlovi:
@dinamic one test is failing (might be related to PHPUnit's update to v6 so we can use PHPStan's plugin):

```
There was 1 failure:

1) FOS\OAuthServerBundle\Tests\Form\Handler\AuthorizeFormHandlerTest::testProcessWillHandleRequestOnPostAndWillProcessDataIfFormIsValid
Failed asserting that Array &0 (
    'client_id' => Binary String: 0x314a761dc55461a75569
    'response_type' => '�V�R�J�ߐ'
    'redirect_uri' => 'U@�Z����5Y'
    'state' => '�qz����&BW'
    'scope' => Binary String: 0x6a00d193d5427e821133
) is identical to Array &0 ().
```

If you could help me out here, that would be great-

dkarlovi:
I'll just mark it as skipped until we figure out why it's failing, don't want a single test block us from moving forward here as I've incorporated PHPStan's PHPUnit extension.

dinamic:
Let me have a look.

dinamic:
Phpunit v6 does not seem to clean the global scope after each test, that's why the test is failing.

You can workaround this by putting `$_GET = [];` in `setUp()`.

Here's what the official site says about this:

```
Prior to version 6, by default, PHPUnit ran your tests in a way where changes to global and super-global variables ($GLOBALS, $_ENV, $_POST, $_GET, $_COOKIE, $_SERVER, $_FILES, $_REQUEST) do not affect other tests.

As of version 6, PHPUnit does not perform this backup and restore operation for global and super-global variables by default anymore. It can be activated by using the --globals-backup option or setting backupGlobals="true" in the XML configuration file.

By using the --static-backup option or setting backupStaticAttributes="true" in the XML configuration file, this isolation can be extended to static attributes of classes.
```

dkarlovi:
Good find, thanks, it should pass now, I'll merge as-is if you don't mind (can look over the changes if you have the time).

Tests/Security/Authentication/Provider/OAuthProviderTest.php
```diff
@@ -35,7 +35,7 @@ class OAuthProviderTest extends \PHPUnit\Framework\TestCase
     protected $userProvider;
 
     /**
-     * @var \PHPUnit_Framework_MockObject_MockObject|OAuthProvider
+     * @var OAuthProvider
```
dinamic: Good catch.

dinamic:
Changes in this PR make sense to me. Thanks for catching up 👍

Tests/Security/Authentication/Provider/OAuthProviderTest.php
dkarlovi: I'd like to take credit, but it was PHPStan. :laughing:

composer.json
```diff
@@ -17,26 +17,29 @@
     ],
     "require": {
         "php": "^7.1",
+        "alcaeus/mongo-php-adapter": "^1.1",
```
dkarlovi: This 100% didn't show up in the PR before I merged it. :see_no_evil:

dinamic: Maybe it got there while you were testing my proposed fix for phpunit v6.

dkarlovi: I've hot-fixed it on master.

dinamic: Don't you actually need that in order to use ODM now? We require PHP 7.1 which doesn't support the dated mongo extension anyway.

dkarlovi: Yes, but we're only suggesting ODM bundle, not requiring it. It's up to the end-user to install the ODM bundle and, by which, this adapter.

dinamic: Yea, you are right. Even if the end-user requires that workaround to use ODM, this is not the correct bundle to take care of the problem.

dkarlovi: If you're not using Mongo ODM, no reason to have this installed.

stof: yeah, this should be moved to the `require-dev` section

composer.json
```diff
@@ -17,26 +17,29 @@
...
+        "phpstan/phpstan-shim": "~0.9",
+        "phpunit/phpunit": "~5.0|~6.0",
+        "propel/propel1": "~1.6",
+        "symfony/class-loader": "~3.0|~4.0",
+        "symfony/console": "~3.0|~4.0",
+        "symfony/form": "~3.0|~4.0",
+        "symfony/phpunit-bridge": "~3.0|~4.0",
+        "symfony/templating": "~3.0|~4.0",
+        "symfony/yaml": "~3.0|~4.0",
+        "willdurand/propel-typehintable-behavior": "~1.0"
```
stof: that change is wrong. You are changing the lowest bound of the requirement

dkarlovi: Thanks @stof, nice catch, will fix it.

PR #517, opened 2018-01-15
Merge 1.6 to master

dinamic:


dinamic:
I am not sure why these two files are shown as conflicting. I have merged them manually using SourceTree and using the website with the `Resolve conflicts` button.

It looks like a bug with Github.

dkarlovi:
Nah, you've merged something incorrectly. If you look at bundle file, the bundle file and its test changed places.

dinamic:
@dkarlovi, you are correct. Should be fixed now. I've reviewed the files and the changes seem fine.

The build does not pass, however, because 1.6 and master have different phpcs rules. Maybe we need to sync them.

dkarlovi:
You can run `make cs` on your branch before committing.

dinamic:
I have applied the phpcs fixes. This wasn't a sound idea IMO, because it would mean we have to make `make cs` per every merge from 1.6 to the master and things will get conflicted over and over again as time goes by.

Now, some other things seem to be broken. I am not sure when I will have time to fix them.

dkarlovi:
@dinamic 1.x will get no further updates at all, we'll almost never need to merge 1.x to master. When we get to functional SF4 support, we'll release and move all development effort to 2.x

dkarlovi:
Your issue here is very simple, PHPStan doesn't know of what types the mocks are. There's an extension for that, but the easiest is to just add a tag above the instance to explain it.

Example:

```php
     * @return \PHPUnit_Framework_MockObject_MockObject|MyClass
     */
    private function mockMyClass(): MyClass
    {
        $service = $this->getMockBuilder(MyClass::class)
            ->disableOriginalConstructor()
            ->getMock();

        return $service;
    }
```

dinamic:
I'm afraid this won't work.

I have tried multiple fixes to no avail.

The errors that phpstan shows at the moment are as follows:

```
 ------ ------------------------------------------------------------------------------------
  Line   Tests/Form/Handler/AuthorizeFormHandlerTest.php
 ------ ------------------------------------------------------------------------------------
  80     Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$query.
  81     Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$request.
 ------ ------------------------------------------------------------------------------------

 ------ ------------------------------------------------------------------------------------
  Line   Tests/Controller/AuthorizeControllerTest.php
 ------ ------------------------------------------------------------------------------------
  200    Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$query.
  205    Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$request.
 ------ ------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------
  Line   Tests/DependencyInjection/Security/Factory/OAuthFactoryTest.php
 ------ -------------------------------------------------------------------------------------
  152    Instantiated class Symfony\Component\DependencyInjection\ChildDefinition not found.
  156    Instantiated class Symfony\Component\DependencyInjection\ChildDefinition not found.
 ------ -------------------------------------------------------------------------------------
```


`Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$query.`

This clearly comes from mocking the ParameterBag in the Request object. The Request object is there and the phpdoc states the type of the object clearly. I've tried changing the places of the classes, but it does not seem to matter - phpstan won't shut up about it.

`Access to an undefined property PHPUnit_Framework_MockObject_MockObject::$request.`

This comes from the Request object again. Same explanation as before.

`Instantiated class Symfony\Component\DependencyInjection\ChildDefinition not found.`

This comes from the `if` condition where we check what is available - is it the ChildDefinition or DefinitionDecorator. I'm not sure how to circumvent this error message. Tried setting the method to `private`, but it didn't work.

I would appreciate any advice as how to make phpstan happy.

dkarlovi:
First, you don't need to actually commit and push to see PHPStan's output, you can run `make phpstan` locally before commiting. Idea is to run `make` before commiting overall, that allows all the checks to run locally for you exactly the same way they run on CI.

About the errors, it's actually a problem the MockObject doesn't contain those, you should denote the params as union, not intersection.

Use `PHPUnit_Framework_MockObject_MockObject&Request` instead of `PHPUnit_Framework_MockObject_MockObject|Request`. As I said, there's a plugin for PHPStan which does this for you, but we'd need to enable it on the project.

dinamic:
Hello @dkarlovi! Thank you for the comment.

Yes, I know I can run `make` locally. Thank you for pointing that anyway.

> Use PHPUnit_Framework_MockObject_MockObject&Request instead of PHPUnit_Framework_MockObject_MockObject|Request. As I said, there's a plugin for PHPStan which does this for you, but we'd need to enable it on the project.

I wasn't aware such notation exists. Using it, however, does not fix the problem. I would be happy if you guide me towards a path that you have personally tested already. I hate to be the one taking comments, testing the proposed solution and respond with 'it does not work'.

I could be that my local instance is somehow now detecting the phpdoc changes, so you can have a look and verify the proposed `union` solution does not make a difference and error message is still there.

As I understand the situation, we have 3 problems to resolve. We are, supposedly, able to resolve two of them using a plugin. How about the last problem?

dkarlovi:
@dinamic I realize this is frustrating, but having tools like this is the only way forward, especially if you have no prior experience with the tools.

You can learn more on union / intersection types in [the blog post by PHPStan's author](https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb). I suggest acquainting yourself with PHPStan as it's an up&coming tool which raises the quality of PHP codebases significantly, I personally use it on all projects I'm working on and it's also seeing adoption on large OSS projects elsewhere.

> How about the last problem?

If you're talking about:

```
 Instantiated class                                                    
         Symfony\Component\DependencyInjection\DefinitionDecorator not found. 
```

that was worked-around by me already in https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/commit/31ff637e512efebafd0916ee6f409b343b409d8e, you can re-apply the solution if necessary.

dinamic:
Thank you for the reference to the article and for the example on how to workaround the issue with the missing class name.

I believe I managed to workaround both issues. Seems like there's a problem with how phpstan works with class members. It does not seem to take into account the phpdoc, rather than the return type of the method that was used to forge the variable value. I shall post an issue at the tracker, so the author could revise.

I hope we will have a green build now :)

dinamic:
We should really squash those commits. Can't Github do it while merging the PR?

dinamic:
On another note, it has been suggested to me to add `phpstan/phpstan-phpunit` as a dev dependency. What do you think about it?

dkarlovi:
> Can't Github do it while merging the PR?

Yes, but that would make next merge from 1.6 much harder as we can expect conflicts from now on. You can [squash your fixes on this merge branch into a single commit](https://stackoverflow.com/questions/2563632/how-can-i-merge-two-commits-into-one), all after https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/517/commits/7979afc720f61b7c72e431f1400de52bd7e68c02 should be OK.

> phpstan/phpstan-phpunit as a dev dependency. What do you think about it?

Definitely, but not on this PR. We're using PHPUnit, PHPStan works better with the extension.

dkarlovi:
BTW thanks again for your effort here, your effort with PHPStan will pay off long-term. 👍 I'm running it in much stricter mode on a large Symfony app I'm building for a client, it keeps finding very obscure and nuanced bugs for me, it's great.

dinamic:
I will re-create the branch and will create another PR. Github doesn't seem to allow changing a PR's branch after the PR has been created already.

Issue #515, opened 2018-01-14
Can't create clients on Windows, uniqueid error


novacp:
When trying to create a client on Windows (command or programatically), uniqueid() for generating a randomToken within FOS\OAuthServerBundle\Util\Random returns an error: 

"Type error: uniqid() expects parameter 1 to be string, integer given in C:\xampp\htdocs\ms-bp-sf4\vendor\friendsofsymfony\oauth-server-bundle\Util\Random.php:31"

dkarlovi:
Hello, thank you for reporting this.

Two things to note:

1. our test case should have caught that (/cc @dinamic WDYT?)
2. you shouldn't be using a branch in development (such as master), it will probably break on you a lot until we add more tests, use a released version

dinamic:
@dkarlovi, we don't have a test coverage for this file.

dinamic:
#518 should fix this.

novacp:
Yes seems legit, but could you merge this please into version 1.6 too? Because I need sf4 and it seems as if 1.6@dev supports it.

Best regards and thanks for your fast response.

PR #518, opened 2018-01-15
Using random_bytes() to generate a random token

dinamic:
Changes:

- No longer having different unique hash strategies across platforms - we are now using random_bytes(). It has been introduced in PHP 7.x, but has been available as a user-land polyfill for a long time.
- added unit test 

Fixes:
- #515 

Added dependencies:
- `paragonie/random_compat` - `random_bytes()` and `random_int()` PHP 5.x polyfill that is already required by `symfony/polyfill-php70`.
- `php-mock/php-mock-phpunit` - useful for mocking php functions in a given namespace. https://github.com/php-mock/php-mock-phpunit

### DIFF
--- a/Tests/Util/RandomTest.php
+++ b/Tests/Util/RandomTest.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace FOS\OAuthServerBundle\Tests\Util;
+
+use FOS\OAuthServerBundle\Util\Random;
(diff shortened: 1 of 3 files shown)


dinamic:
This can actually go to 1.5.

dinamic:
Ping @dkarlovi

dkarlovi:
Thank you @dinamic

PR #519, opened 2018-01-15
Allow template overrides in sf >= 3.4

andythorne:
Without the `@` symbol, symfony templating will look in the `__main__` namespace, rather than `FOSOAuthServerBundle`.

### DIFF
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -195,7 +195,7 @@ class AuthorizeController implements ContainerAwareInterface
         }
 
         return $this->templating->renderResponse(
-            'FOSOAuthServerBundle:Authorize:authorize.html.'.$this->templateEngineType,
+            '@FOSOAuthServerBundle:Authorize:authorize.html.'.$this->templateEngineType,
             [
                 'form' => $form->createView(),
                 'client' => $this->getClient(),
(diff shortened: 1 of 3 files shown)


dinamic:
@andythorne, I'm not in the loop. Would this be compatible with sf <3.4 as well?

andythorne:
Yes,  I believe so.

andythorne:
According to the 3.3 docs, the @ notation is valid: http://symfony.com/doc/3.3/templating.html#referencing-templates-in-a-bundle

dkarlovi:
Could we have a functional test for that?

andythorne:
Master only supports >=3.0, and 3.0 supports bundle namespaces: https://symfony.com/doc/3.0/templating/namespaced_paths.html

I don't think adding a test for this is necessary, as that would just be testing twig...

dkarlovi:
@andythorne no, we'd be testing **our usage of Twig** with specific Symfony / Twig versions, there's a difference.

Reading the docs, you seem to be right, it's supported with 3.0+. I'd still like a single test (which ends up running on all supported Symfony versions, mind you), just to prove `render()` doesn't throw.

andythorne:
I see what your saying, but our usage of twig is as defined in the docs, which are backed by unit tests under symfony.

Hence, I don't thing we need a functional test (which actually seem to be broken atm anyway, and are marked incomplete). 

I see we already have a unit test for checking the render()  method arguments. Worryingly, the test passed... I'll look into why that is.

andythorne:
@dkarlovi rebase and updated the tests to account for the new syntax.

dkarlovi:
Thanks @andythorne

Issue #520, opened 2018-01-16
Make fos_oauth_server.server public


SebScoFr:
Sometimes it is useful to have the ability to manually grant an access token from the OAuth server.
```php
$oauth = $this->container->get('fos_oauth_server.server');
$response = $oauth->grantAccessToken($request);
```
Unfortunately, this doesn't work out of the box anymore since the service is private by default. Would it be ok to make it public?

Also, @dkarlovi can you merge PRs from the **oauth2-php** repo? https://github.com/FriendsOfSymfony/oauth2-php/pull/105 is really needed with SF4.

SebScoFr:
Created PR #522

dkarlovi:
@SebScoFr why not just inject it? That's the whole point why services got private by default, it promotes injecting instead of fetching (by which you're hiding the dependency).

SebScoFr:
@dkarlovi I have a service that already have access to the container, so it seemed a bit strange to me to inject the oauth server service alongside the container, when I could just grab it from the container, if that makes sense. But if there is no way around it I'll end up doing that.

dkarlovi:
@SebScoFr you should inject the container too, it's also an anti-pattern. You should inject your direct dependencies. If you need some services which cannot be injected (there are obscure situations), use [Service Locators](https://symfony.com/blog/new-in-symfony-3-3-service-locators).

PR #522, opened 2018-01-16
Update oauth.xml

SebScoFr:


dkarlovi:
See discussion in #520 why we're not doing this.

Issue #529, opened 2018-01-31
Symfony 4 - Non-public service should be made public


cvele:
Since all services are made private by default in Symfony 4, the same applies to `fos_oauth_server.client_manager.default`, which we should be able to fetch publicly from container.

dkarlovi:
Duplicate of #520.

PR #530, opened 2018-01-31
Make service client manager public

cvele:
`fos_oauth_server.client_manager.default` should be public by default, Symfony 4 #529


cvele:
Closing due to explanation in #520

PR #531, opened 2018-02-02
Fix compatibility for PHP > 7.0

DoulyCreations:


dkarlovi:
We will not be accepting this, we're targeting strict types and nullables are a must. Symfony 4 requires 7.1.3+, we should follow.

PR #532, opened 2018-02-03
Update AuthorizeController.php

Romaxx:
in dev mode (SF 3.4.4) Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher is passed to arg 10 constructor whereas Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher is passed in prod mode, so Interface Class Symfony\Component\EventDispatcher\EventDispatcherInterface is required


dkarlovi:
You should update all the references in the file, not just the constructor.

dkarlovi:
Fixed in #533

PR #533, opened 2018-02-06
Use EventDispatcherInterface

andythorne:
Using `EventDispatcher` throws an exception on sf >= 3.4 in when the debugger is enabled, as the debugger wraps the `EventDispatcher` with `Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher`.


dkarlovi:
Thanks @andythorne

PR #534, opened 2018-02-06
Fixing unregistered commands

SebScoFr:


dkarlovi:
Thank you @SebScoFr

Command/BaseCommand.php
```diff
@@ -0,0 +1,40 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\OAuthServerBundle\Command;
+
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Console\Command\Command;
+
+class BaseCommand extends Command implements ContainerAwareInterface
```
dinamic: This class causes a problem for #536.

How is this class different than the one used before - `ContainerAwareCommand`? I mean, what's the benefit of having one more file to maintain. It seems to be doing the exact same thing.

dkarlovi: You're right, I've merged this too quickly. :+1: 

We should work on getting actual functional tests running, ie. having a minimal Symfony app setup and the bundle registered there.

SebScoFr: The main idea is to have a shared base class that provide the container, rather than having to inject it in every single command. I just find it cleaner.
The main purpose of this PR though isn't that, but the changes in *Resources/config/oauth.xml* so that the commands get registered.

PR #535, opened 2018-02-07
Raise PHPStan's level from 3 to 4

dkarlovi:


Form/Model/Authorize.php
```diff
@@ -49,7 +49,7 @@ class Authorize
     public $scope;
 
     /**
-     * @param bool  $accepted
+     * @param mixed $accepted
```
dkarlovi: This is probably wrong, need to look at it again.

PR #539, opened 2018-02-15
fix usages of deprecated constants

### LINKED ISSUE #528: Problem initializing
`OAuth2ServerException::getHttpResponse` is attempting to pass the result of `OAuth2ServerException::getHttpCode()` to Symfony Response object - which results in a ` Uncaught PHP Exception ErrorException: "Notice: A non well formed numeric value encountered"`

**Quick debug:**

`Symfony\Component\HttpFoundation\Response::__construct` in it's method signature requires http status argument to be integer. In the case of this bundle http status codes are held in `OAuth2\OAuth2` as constants with string assigned values eg. `HTTP_UNAUTHORIZED = '401 Unauthorized'`.

I'm using Symfony HttpFoundation at version 4.0 and OAuth bundle is fetched from master (tried 1.6 branch too).

Would gladly send a PR that would set those constant codes to pure integers, but I'm not sure if there will be any other repercussions (are they used anywhere else?).

tomasjav:
fixes #528

### DIFF
--- a/Security/Authentication/Provider/OAuthProvider.php
+++ b/Security/Authentication/Provider/OAuthProvider.php
@@ -12,6 +12,7 @@
 namespace FOS\OAuthServerBundle\Security\Authentication\Provider;
 
 use FOS\OAuthServerBundle\Security\Authentication\Token\OAuthToken;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Provider\AuthenticationProviderInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
@@ -74,7 +75,7 @@ class OAuthProvider implements AuthenticationProviderInterface
                     try {
                         $this->userChecker->checkPreAuth($user);
                     } catch (AccountStatusException $e) {
-                        throw new OAuth2AuthenticateException(OAuth2::HTTP_UNAUTHORIZED,
+                        throw new OAuth2AuthenticateException(Response::HTTP_UNAUTHORIZED,
                             OAuth2::TOKEN_TYPE_BEARER,
                             $this->serverService->getVariable(OAuth2::CONFIG_WWW_REALM),
                             'access_denied',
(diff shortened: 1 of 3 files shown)


GuilhemN:
Thank you @tomasjav!

dkarlovi:
@GuilhemN it's great to see you back here reviewing and merging! :+1:

GuilhemN:
I won't be able to do this much unfortunately. That was a simple issue here but quite bothering for users so I took some time for it :)

I'm not sure I merged this PR in the right branch btw, a small explanation in the readme about how we use banches would be useful imo :)

dkarlovi:
Oh, I thought I did a bad enough job you decided enough was enough and got back. :smile: 

`1.5` is as you've left it, it won't be touched anymore.

`1.6` is there for us to get Symfony 4 support and try to fix most crucial bugs (so you did merge correctly), it will be deprecated as soon as Symfony 2 is.

`master` will become 2.0 and will be the only place where any new development should be done.

GuilhemN:
No you've done a great job dont worry 😃 in fact I intervene in most fos projects from time to time when someone pings me.

> master will become 2.0 and will be the only place where any new development should be done.

Ok thanks, I'll merge the commit in master too then :)

stof:
@dkarlovi I don't actually agree with your description here. any new feature which can be implemented in a backward compatible way should rather go in a 1.x release instead of only 2.0, to allow easier migration (similar to Symfony 3.4 getting all new features, not 4.0)

dkarlovi:
@stof I see your point, we're implementing **some** features to 1.6 (see Milestones), but bulk of the work will be incompatible anyway and I don't see backporting it back, it would require a significant amount of work (with 3 major versions of Symfony being supported) which this project can't afford IMO.

We're deprecating stuff in 1.6 as they're changed for 2.x, that should be enough?

Of course, if you disagree, feel free to correct me as you're the senior officer here. :smile:

dkarlovi:
Of course, we could opt for 1.7 instead of 2.0 and just drop support for Symfony 2 there, you might have a point.

GuilhemN:
I couldn't find a BC break in master so I'd rather go for 1.7.

PR #551, opened 2018-03-22
Fix error DependancyInjection Exception

ThomAille:
I changed alias to class because Symfony retraces an Injection dependency error :
[Symfony\Component\DependencyInjection\Exception\RuntimeException]                                                                                                                                      
  The definition for "fos\oauthserverbundle\controller\tokencontroller" has no class. If you intend to inject this service dynamically at runtime, please mark it as synthetic=true. If this is an abstr  
  act definition solely used by child definitions, please add abstract=true, otherwise specify a class to get rid of this error.

### DIFF
--- a/Resources/config/oauth.xml
+++ b/Resources/config/oauth.xml
@@ -27,7 +27,7 @@
             <argument type="service" id="fos_oauth_server.server" />
         </service>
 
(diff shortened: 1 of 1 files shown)


dkarlovi:
@ThomAille thank you.

PR #552, opened 2018-03-26
Allow array in supported_scopes option

jdeniau:
Currently, if you have an important number of scopes, you must set them into a one-line config option, which come to be really hard to read:

```yaml
fos_oauth_server:
      service:
          options:
              supported_scopes: scope1 scope2 scope3 scope4 scope5 scope6
```

This PR allow using an array:
```yaml
fos_oauth_server:
      service:
          options:
              supported_scopes:
                - scope1
                - scope2
                - scope3
                - scope4
                - scope5
                - scope6
```

It still works with a string if you have a small number of scopes
```yaml
fos_oauth_server:
      service:
          options:
              supported_scopes: scope1 scope2
```

### DIFF
--- a/Controller/AuthorizeController.php
+++ b/Controller/AuthorizeController.php
@@ -249,7 +249,7 @@ class AuthorizeController implements ContainerAwareInterface
     }
 
     /**
-     * @return ClientInterface.
+     * @return ClientInterface
      */
     protected function getClient()
     {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -104,7 +104,7 @@ class Configuration implements ConfigurationInterface
                             ->arrayNode('options')
                                 ->useAttributeAsKey('key')
                                 ->treatNullLike([])
-                                ->prototype('scalar')->end()
+                                ->prototype('variable')->end()
                             ->end()
                         ->end()
                     ->end()
--- a/DependencyInjection/FOSOAuthServerExtension.php
+++ b/DependencyInjection/FOSOAuthServerExtension.php
@@ -52,8 +52,8 @@ class FOSOAuthServerExtension extends Extension
         }
 
         $options = $config['service']['options'];
-        if (!empty($options['supported_scopes'])) {
-            $options['supported_scopes'] = str_replace("\n", ' ', $options['supported_scopes']);
+        if (is_array($options['supported_scopes'] ?? null)) {
+            $options['supported_scopes'] = implode(' ', $options['supported_scopes']);
         }
         $container->setParameter('fos_oauth_server.server.options', $options);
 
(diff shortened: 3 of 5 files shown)


Controller/AuthorizeController.php
```diff
@@ -249,7 +249,7 @@ protected function getRedirectionUrl(UserInterface $user)
     }
 
     /**
-     * @return ClientInterface.
```
jdeniau: marked as error by php-cs-fixed  ¯\_(ツ)_/¯

dkarlovi:
I don't like this approach. I'd rather we switch to an array in the config (deprecating the string first, of course).

jdeniau:
@dkarlovi move to an array definition. I had to change `prototype('scalar')` by `prototype('variable')` though because I did not manage to make a scalar prototype except for the `supported_scope` attribute but it may be more logical because the options parameters are managed by the oauth2-php library directly and the library might accept any options (accepted values [are `mixed` by the way](https://github.com/FriendsOfSymfony/oauth2-php/blob/a41fef63f81ef2ef632350a6c7dc66d15baa9240/lib/OAuth2.php#L451)).

About the string deprecation vs array, it does now works with both, and I think it's a good idea to allow string if you only have quite a few scopes.
If you want to enforce arrays though, and deprecate the string parameter "the symfony way", I will be stucked like I said before with the configuration. (I can throw a deprecation in the FOSOauthServerExtension.php but it's not the best option as Symfony [provide a way to deprecate configuration parameter](http://symfony.com/doc/current/components/config/definition.html#deprecating-the-option))

dkarlovi:
@jdeniau you're right, looking at this, there's no need to deprecate the string notation, array notation is just a DX improvement. This looks good to me as is. :+1: 

Good to merge?

jdeniau:
@dkarlovi of course ! It's done

dkarlovi:
LGTM, but I just remembered (sorry to be a nag): do you think we should check for a space in individual array items?

So, for example

```yaml
scopes:
- foo bar
- bar
```

will become `(foo) (bar) (bar)`, and not `(foo bar) (bar)` as one might expect.

jdeniau:
Hmmm... sorry I don't think I understand:

The only manipulation is an `implode(' ')`, so no matter how you put your scopes, you can write:
```yaml
- foo
- bar
- bar
```
```yaml
foo bar bar
```
or
```yaml
- foo bar
- bar
```

the output will still be: `foo bar bar`

jdeniau:
That "could" be interesting by the way if you want to do something like:
```yaml
scopes:
  - foo:write foo:read
  - bar:write bar:read
  - baz:write baz:read
```

dkarlovi:
> the output will still be: foo bar bar

That's the problem, I feel. We probably shouldn't allow a space in array items at all since it's a separator.

jdeniau:
In that case I can throw a [Symfony Config InvalidConfigurationException](http://api.symfony.com/3.4/Symfony/Component/Config/Definition/Exception/InvalidConfigurationException.html) if one of the item contains a space or a tab

dkarlovi:
We should do that and, other than that, I see no reason not to merge here. :+1:

dkarlovi:
@jdeniau awesome work, TYVM! \o/

PR #554, opened 2018-04-03
The authorize form should be used only if authorize option true.

makasim:


dkarlovi:
Why haven't our tests caught this?

makasim:
I am build app on top of Symfony Flex, there is no forms and templating in my setup, things breaks here and there (:

makasim:
Added tests

dkarlovi:
This was previously done in #466, but since you've added tests, we should merge this one instead.

PR #558, opened 2018-04-06
update ODM part of example: add user property in AccessToken

wingzero0:
update ODM part of example: add user property in AccessToken, RefressToken, AuthCode. 

Missing it will cause mongodb not saving the referenced user, and in the authenication step, it will not able to get back user.

Also replace the mapping with annotation for reading reasier.


alcaeus:
#483 already does some improvements in the docs for ODM, even though we weren't quite sure whether we'd want to use annotations or XML.

PR #566, opened 2018-04-17
Mark a client as trusted

fxbt:
I'm using OAuth2 for internal service authentication and the authorization form is not needed.

This PR skips the authorization screen if the client is marked as trusted.


AKonnov:
I agree with the @fxbt . This is useful functionality.

PR #572, opened 2018-04-23
infra: Enable Infection mutation framework integration

dkarlovi:


infection.json.dist
```diff
@@ -0,0 +1,15 @@
...
+        ],
+        "exclude": [
+            "Tests",
+            "vendor"
+        ]
+    },
+    "logs": {
+        "text": "build/infection-log.txt"
+    }
+}
```
dkarlovi: Why U NO EOL?!

PR #574, opened 2018-04-26
Don't get repository at construction time, it could be not loaded yet!

fullbl:
I ran into an issue when trying to load some repositories through compiler pass.
At Doctrine\Bundle\DoctrineBundle\Repository\ContainerRepositoryFactory->getRepository(), it tried to load repository of my User Entity.

I changed the TokenManager, so it doesn't try to get it too soon!

### DIFF
--- a/Entity/TokenManager.php
+++ b/Entity/TokenManager.php
@@ -26,26 +26,29 @@ class TokenManager extends BaseTokenManager
     protected $em;
 
     /**
-     * @var EntityRepository
-     */
-    protected $repository;
-
-    /**
      * @var string
      */
     protected $class;
 
     public function __construct(EntityManagerInterface $em, $class)
     {
-        // NOTE: bug in Doctrine, hinting EntityRepository|ObjectRepository when only EntityRepository is expected
-        /** @var EntityRepository $repository */
-        $repository = $em->getRepository($class);
-
         $this->em = $em;
-        $this->repository = $repository;
         $this->class = $class;
     }
 
+    /**
+     * retrocompatibility with old $repository property
+     * @param $name
+     * @return mixed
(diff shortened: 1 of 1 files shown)


fullbl:
After some more researches, I discovered that could be my installation fault. I'll leave this PR open, since it could be good to remove that call from constructor!

Entity/TokenManager.php
```diff
@@ -25,27 +25,30 @@ class TokenManager extends BaseTokenManager
...
-        $this->repository = $repository;
         $this->class = $class;
     }
 
+    /**
+     * retrocompatibility with old $repository property.
+     * @param $name
+     * @return mixed
+     */
+    public function __get($name)
```
stof: should not be public to provide BC with a *protected* property.

and it should trigger a deprecation warning

Entity/TokenManager.php
```diff
@@ -25,27 +25,30 @@ class TokenManager extends BaseTokenManager
...
+    {
+        if ('repository' === $name) {
+            return $this->getRepository();
+        }
+
+        return $this->$name;
```
stof: should not allow access. This gives public access to private properties

Entity/TokenManager.php
fullbl: __get can only be public... I can't think a method to provide retrocompatibility for private property.

Entity/TokenManager.php
fullbl: Removed!

GuilhemN:
👍 for me, phpstan is broken however.

alcaeus:
Note: this also applies to other repositories, not just EntityRepositories.

PR #590, opened 2018-08-02
POC Introspection endpoint RFC 7662

jdeniau:
This is a work in progress but I would like your opinion about implementing the [RFC 7662:  OAuth 2.0 Token Introspection](https://tools.ietf.org/html/rfc7662).

There are some missing things, some code to move but it opens the discussion. 📢

A "blocking" point for me that I see is how to handle security.
My first idea was to use the oauth2 work-flow and add a list of "allowed" client for the introspection endpoint, but it seems kind of hacky :thinking:


jdeniau:
ping @dkarlovi maybe ?

PR #592, opened 2018-08-29
SF 4.1: Referencing controllers with a single colon is deprecated 

lisfox1:
Fixing deprecated notice for Symfony 4.1: 
`Referencing controllers with a single colon is deprecated since Symfony 4.1`

### DIFF
--- a/Resources/config/routing/authorize.xml
+++ b/Resources/config/routing/authorize.xml
@@ -5,7 +5,7 @@
     xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
 
     <route id="fos_oauth_server_authorize" path="/oauth/v2/auth" methods="GET POST">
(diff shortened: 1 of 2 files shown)


stof:
I'm rejecting this for the same reason than in all other FOS bundles for now: the change is dropping support for all versions of Symfony older than 4.1, which is not an option.

dkarlovi:
This should be done the same way I've seen elsewhere: ship additional routing files, like `authorize_symfony40.xml` or similar.

lisfox1:
@stof Would you consider the PR if I used the idea of having separate files for the SF4?

stof:
but even then, we would have to find a way to make it without BC break (i.e. we cannot just ship another file for older version and update the existing file, as all users would have to update their projects).

I'm still evaluating what would be a transparent approach to this BC layer (I don't want to have projects being responsible for choosing a different routing file, as that would even forbid us to remove the old one in the future)

PR #598, opened 2018-12-06
Fix deprecations from Symfony 4.2

alcaeus:
This fixes the tree builder deprecation that comes up starting with Symfony 4.2.

@dkarlovi unfortunately, the `RandomTest` is broken when using Symfony 3 or later; `runInSeparateProcess` produces a fatal error:
```
Fatal error: Uncaught Error: Using $this when not in object context in /private/var/folders/rn/1wpg9zfn4gbd84dl26xzbfzm0000gn/T/FOSOAuthServerBundle/ContainerHAkME25/getCacheWarmerService.php on line 16

Error: Using $this when not in object context in /private/var/folders/rn/1wpg9zfn4gbd84dl26xzbfzm0000gn/T/FOSOAuthServerBundle/ContainerHAkME25/getCacheWarmerService.php on line 16

Call Stack:
    0.0143     966088   1. {main}() Standard input code:0
    0.3101   27238072   2. require_once('/private/var/folders/rn/1wpg9zfn4gbd84dl26xzbfzm0000gn/T/FOSOAuthServerBundle/ContainerHAkME25/getCacheWarmerService.php') Standard input code:1219
```

Dropping `runInSeparateProcess` drops the fatal but causes an assertion failure in the test. How do you want me to proceed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ matrix:
     - php: 7.2
       env: SYMFONY_VERSION=2.8.* COVERAGE=true
     - php: 7.2
+      env: SYMFONY_VERSION=4.2.* COVERAGE=true
+    - php: 7.2
       env: SYMFONY_VERSION=3.4.* COVERAGE=true
     - php: 7.2
       env: DEPENDENCIES=beta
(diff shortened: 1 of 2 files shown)


alcaeus:
Closing here: with the build for 1.x badly broken I can't be bothered to resurrect it. We'll fix this stuff in 2.0 and try to get that shipped. See #599.

Issue #478, opened 2017-09-01
Enable to specify user_checker in firewall settings


GSokol:
In current realization the only symfony's system `security.user_checker` is provided to `OS\OAuthServerBundle\Security\Authentication\Provider\OAuthProvider`'s container. It's rather trickiy to inject there the custom one for the certain firewall area.

denisWetexx:
Is there any concern in making the fix public?
I had the same problem trying to use the user_checker as proposed in https://symfony.com/doc/current/security/user_checkers.html
The configured user_checker does not get injected. Instead the default user checker is used 
in vendor/friendsofsymfony/oauth-server-bundle/Security/Authentication/Provider/OAuthProvider.php
so checking for disabled users is not possible without the fix for non AdvancedUserInterfaces or more custom authorization checks

deguif:
It's now supported in master branch, see #660

PR #602, opened 2019-01-21
FIX: adds support for using custom user checker

nvbooster:
Injects user checker service specified in firewall configuration to authentication provider injecting same way as it is done in standard symfony providers

Custom user checker service is always created by symfony security based on standard firewall.user_checker configuration option, all we need is to inject it to authentication [provider]

Solves #478


mctdev-marius:
Can you please, try and make it in the master branch. This is a required function for custome user checkers

deguif:
It's now supported in master branch, see #660

PR #617, opened 2019-06-16
Use form_end instead of closing HTML tag by hand

adrienbrignon:
self explanatory


deguif:
Thanks @adrienbrignon 
Let's bring this bundle back to life

PR #631, opened 2019-10-16
Update translation nl

bghosts:


deguif:
Thanks @bghosts
Let's bring this bundle back to life

PR #662, opened 2020-10-03
Fix "referencing controllers with a single colon is deprecated"

GKFX:
This commit is the same as https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/592
However now SF <4.4 are officially not supported this should not break BC.


deguif (approved):

deguif:
Thank you @GKFX

PR #663, opened 2020-10-05
Make CI useful again and add type declarations

GKFX:
This pull request:
  - Adds type declarations to bring down the PHPStan error count
  - Makes the CI not dependant on a PHPStan pass so that it can run through to the unit tests
  - Adds alcaeus/mongo-php-adapter so that `composer install` can run without `ext-mongo`. This could be removed again if doctrine/mongodb-odm is upgraded, as in the elchris fork. 
  - Disables the Propel tests as all the Propel code seems to be missing something in the `FOS\OAuthServerBundle\Propel\om` namespace
  - Updates the PHPStan configuration including the addition of phpstan-symfony.

The first commit is unfortunately rather large. A lot of it comes from running `sed` over `Tests/` to add return type declarations, which inevitably produced a lot of changes.


deguif:
@GKFX really nice, but could you split this PR in to smaller ones, please.

PR #668, opened 2020-10-17
Added default names in the commands for lazy loading

magnusreiss:
We are using the OAuth server bundle in a large application and we have noticed that the commands are always loaded when called from the console. So we would be happy if the change is accepted.

See documentation for lazy loading: https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading
There are no negative effects.

### DIFF
--- a/Command/CleanCommand.php
+++ b/Command/CleanCommand.php
@@ -21,7 +21,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class CleanCommand extends Command
 {
-
     protected static $defaultName = 'fos:oauth-server:clean';
 
     private $accessTokenManager;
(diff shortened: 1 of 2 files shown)


Command/CleanCommand.php
```diff
@@ -45,7 +47,7 @@ protected function configure()
         parent::configure();
 
         $this
-            ->setName('fos:oauth-server:clean')
+            ->setName(self::$defaultName)
```
deguif, line 50: This can be removed as it's automatically done by the base command constructor starting from Symfony 3.4 and we now support 4.4 and upper in master.

Command/CreateClientCommand.php
```diff
@@ -39,7 +41,7 @@ protected function configure()
         parent::configure();
 
         $this
-            ->setName('fos:oauth-server:create-client')
+            ->setName(self::$defaultName)
```
deguif, line 44: This can be removed too.

deguif (commented):
Just minor comments

magnusreiss:
Thanks and done

deguif:
Thanks @magnusreiss

PR #676, opened 2021-03-28
Re-add templating engine

dennisameling:
In #653, the templating engine was removed in favor of Twig. PHP templates [have been deprecated](https://symfony.com/doc/4.4/templating/PHP.html) in Symfony 4.3 and removed in Symfony 5.

However, in https://github.com/mautic/mautic we still heavily rely on PHP templates and are not ready to move to Twig templates just yet, as there's a lot of technical debt involved. Symfony 4.4 [will be supported](https://symfony.com/releases/4.4) until November 2022 and will receive security fixes until November 2023. Would you please consider re-adding the template engine for now? Otherwise we can't get PHP 8.0 support to Mautic as soon as you release v2 of this bundle.

Thanks in advance!


nickveenhof:
Ping? Can we get some help? This is blocking Mautic from moving forward with PHP 8. Happy to add tests or anything to move this forward :)

PR #678, opened 2021-04-01
Simplify OAuthEventClass documentation

jdeniau:
As the `$event->getClient()` method [returns a ClientInterface](https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/Event/AbstractAuthorizationEvent.php#L59), there is no need to check that this value is not null


deguif:
Indeed

Thanks @jdeniau

PR #687, opened 2021-12-14
hidourianis: fix token controller type hinting

HidouriAnis:
We should use Interface type hinting. i have a problem with this type hinting when  I needed to create a decorator for a custom createAccessToken method


deguif (approved):
Could you add a changelog?

HidouriAnis:
@deguif  do you have any idea why tests are 
failed?

